PR #104, opened 2017-02-06
Replace "licence" with "license"

koogawa:
## Issue
-

## Overview (Required)
"licence" and "license" are mixed in project. I think those should be unified.

## Links
- [Licence vs\. License: What’s the Difference? \- Writing Explained](http://writingexplained.org/licence-vs-license-difference)

## Screenshot
Before | After
:--: | :--:
 |


hotchemi:
@koogawa LGTM! Thanks for you dedicated work!

konifar:
Thanks! You are both iOS and Android engineer 💪

PR #105, opened 2017-02-06
grouping licenses in licenses.yml

kgmyshin:
## Issue
-

## Overview (Required)
- grouping licenses in licenses.yml 
  - I think grouping licenses is better than skipping other except main.
- add checklicenses-task in check-task.

## Links
-

I confirmed generated html is not changed.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -297,5 +297,7 @@ task jacocoTestReportDevelop(type: JacocoReport, dependsOn: ['testDevelopDebugUn
     }
 }
 
+tasks['check'].dependsOn checkLicenses
+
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -297,5 +297,7 @@ task jacocoTestReportDevelop(type: JacocoReport, dependsOn: ['testDevelopDebugUn
     }
 }
 
+tasks['check'].dependsOn checkLicenses
```
kgmyshin: I think check-task should do checkLicenes-task.
But you think this is too hard, I remove this line 🎅

konifar:
Good catch!! LGTM 💯

Issue #71, opened 2017-02-03
Add loading view in sessions page


konifar:
## Overview (Required)
Now sessions page has no loading view.
If the data is not cached, it takes about 1 sec to show sessions.

PR #106, opened 2017-02-06
Show progress bar on loading sessons

kobakei:
## Issue
- #71 

## Overview (Required)
- Show progress bar on loading sessions data

## Links
-

## Screenshot
Before | After
:--: | :--:
 | device-2017-02-06-170634

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionsViewModel.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionsViewModel.java
@@ -4,8 +4,11 @@ import com.annimon.stream.Collectors;
 import com.annimon.stream.Stream;
 
 import android.content.Context;
+import android.databinding.BaseObservable;
+import android.databinding.Bindable;
 import android.support.annotation.NonNull;
 import android.text.TextUtils;
+import android.view.View;
 
 import java.util.ArrayList;
 import java.util.Date;
@@ -15,6 +18,7 @@ import java.util.Map;
 
 import javax.inject.Inject;
 
+import io.github.droidkaigi.confsched2017.BR;
 import io.github.droidkaigi.confsched2017.model.MySession;
 import io.github.droidkaigi.confsched2017.model.Room;
 import io.github.droidkaigi.confsched2017.model.Session;
@@ -23,7 +27,7 @@ import io.github.droidkaigi.confsched2017.repository.sessions.SessionsRepository
 import io.github.droidkaigi.confsched2017.util.DateUtil;
 import io.reactivex.Single;
 
-public class SessionsViewModel implements ViewModel {
+public class SessionsViewModel extends BaseObservable implements ViewModel {
 
     private SessionsRepository sessionsRepository;
 
@@ -55,6 +59,8 @@ public class SessionsViewModel implements ViewModel {
 
                     this.rooms = extractRooms(sessions);
                     this.stimes = extractStimes(sessions);
+
+                    notifyPropertyChanged(BR.loadingVisibility);
 
                     List<SessionViewModel> viewModels = Stream.of(sessions)
                             .map(session -> {
@@ -157,5 +163,14 @@ public class SessionsViewModel implements ViewModel {
 
     public List<Date> getStimes() {
         return stimes;
+    }
+
+    @Bindable
+    public int getLoadingVisibility() {
+        if (this.rooms == null) {
+            return View.VISIBLE;
+        } else {
+            return View.GONE;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


konifar:
:eyes:

app/src/main/res/layout/fragment_sessions.xml
```diff
@@ -85,6 +86,21 @@
                 android:textColor="@color/white"
                 />
 
+        <FrameLayout
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                android:background="@color/white"
```
konifar: `background` is not necessary.

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionsViewModel.java
```diff
@@ -56,6 +60,8 @@ public void destroy() {
                     this.rooms = extractRooms(sessions);
                     this.stimes = extractStimes(sessions);
 
+                    notifyPropertyChanged(BR.loadingVisibility);
```
konifar: I think it works well. But I prefer Rx way like this.
http://qiita.com/boohbah/items/e8010730725c54f85a3a

What do you think? Just I want to know your opinion 😃

konifar:
@kobakei Thanks for your second contribution!! Commented!

app/src/main/res/layout/fragment_sessions.xml
kobakei: If the background is undefined, the date text view with empty text and gray background will be shown. Is it ok?
device-2017-02-06-172056

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionsViewModel.java
kobakei: I don't have confidence, but I comment :)

If the project uses data binding, the visibility of views should be controlled by data binding as possible. However, if we use progress dialog instead of progress view in layout xml, it is difficult to control the visibility of dialog only with data binding and the method described in the article may be good.

In this case, I prefer progress bar than progress dialog because we use BottomNavigationView. Even if the session data is not loaded, users should be able to change tab.

app/src/main/res/layout/fragment_sessions.xml
konifar: Ah, I see. Make sense

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionsViewModel.java
konifar: Yep, I agree to update the visibility via DataBinding and to use progress bar!
My suggestion is to use `doOnSubscribe()`.

```xml
return Single.zip(sessionsRepository.findAll(languageId),
                mySessionsRepository.findAll(),
                (sessions, mySessions) -> {
                    final Map<Integer, MySession> mySessionMap = new LinkedHashMap<>();
                    for (MySession mySession : mySessions) {
                        mySessionMap.put(mySession.session.id, mySession);
                    }

                    this.rooms = extractRooms(sessions);
                    this.stimes = extractStimes(sessions);

                    List<SessionViewModel> viewModels = Stream.of(sessions)
                            .map(session -> {
                                boolean isMySession = mySessionMap.containsKey(session.id);
                                return new SessionViewModel(session, context, rooms.size(), isMySession);
                            })
                            .collect(Collectors.toList());
                    return adjustViewModels(viewModels, context);
                })
                .doOnSuccess(sessionViewModels -> {
                    setLoadingVisibile(false);
                })
                .doOnError(throwable -> {
                    setLoadingVisibile(false);
                })
                .doOnSubscribe(disposable -> {
                    setLoadingVisibile(true);
                })
```

kobakei: ah i see.
I think your code is easy to understand.
I will fix it.

konifar: Thanks!! 🙇

app/src/main/res/layout/fragment_sessions.xml
kobakei: I found an issue to update visibility in doOnSuccess/doOnError.
doOnSuccess and doOnError run on not-UI thread therefore notifyPropertyChanged causes CalledFromWrongThreadException.
Of course I can notify with Handler... what should I do?

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionsViewModel.java
kobakei: I found an issue to update visibility in doOnSuccess/doOnError.
doOnSuccess and doOnError run on not-UI thread therefore notifyPropertyChanged causes CalledFromWrongThreadException.
Of course I can notify with Handler... what should I do?
Using Handler just for notifyPropertyChange may be too much.

konifar: Ah, I see.
Okay, I don't have strong opinion. So it's okay the way you have fixed 👍

konifar:
LGTM! 💯

PR #107, opened 2017-02-06
Delete modules.xml

gen0083:
## Issue
I think that `.idea/modules.xml` is not necessary to put under version control.
When I cloned repository and renamed it, I had to update `.idea/modules.xml` file.

This file should depend on the environment, I think that it is not necessary to put under version control.

## Overview (Required)
- 

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


gfx:
+1

konifar:
:eyes:

konifar:
LGTM!! Good catch! Thanks! 👏

PR #109, opened 2017-02-06
Fix crash when language resource is null 

nshiba:
## Issue
-

## Overview (Required)
Like an after party session, If topic color is null it is crash. 
So I set the default value.

and

if topic or room is null, I set View.GONE to session tag.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
@@ -38,16 +38,16 @@ public class SessionDetailViewModel extends BaseObservable implements ViewModel 
     private String sessionTitle;
 
     @ColorRes
-    private int sessionVividColorResId;
+    private int sessionVividColorResId = TopicColor.NONE.vividColorResId;
 
     @ColorRes
-    private int sessionPaleColorResId;
+    private int sessionPaleColorResId = TopicColor.NONE.paleColorResId;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
```diff
@@ -38,16 +38,16 @@
     private String sessionTitle;
 
     @ColorRes
-    private int sessionVividColorResId;
+    private int sessionVividColorResId = TopicColor.NONE.vividColorResId;
```
konifar: > If topic color is null it is crash.

I guess it doesn't occur crash because `TopicColor.from` method returns NONE when topic is null.

```java
    public static TopicColor from(@Nullable Topic topic) {
        if (topic == null) {
            return NONE;
```

So maybe it's not necessary to set default value.

nshiba: Sorry...
It was when the language was not set.

So, if the language is null, I need to set the default value (`R.string.lang_en`)

konifar: Understood! Thanks!

nshiba:
fixed f92eed8

konifar:
Thank you for ! LGTM 💯

Issue #117, opened 2017-02-06
Crash App, Resources NotFoundException


futabooo:
## Overview (Required)
- if session's lang is null SessionDetailFragment can not open

## Links
- 

## Screenshot
```
02-06 21:12:21.832 26463-26463/io.github.droidkaigi.confsched2017.develop.debug E/AndroidRuntime: FATAL EXCEPTION: main
  Process: io.github.droidkaigi.confsched2017.develop.debug, PID: 26463
  android.content.res.Resources$NotFoundException: String resource ID #0x0
    at android.content.res.Resources.getText(Resources.java:312)
    at android.widget.TextView.setText(TextView.java:4422)
    at io.github.droidkaigi.confsched2017.databinding.FragmentSessionDetailBinding.executeBindings(FragmentSessionDetailBinding.java:263)
    at android.databinding.ViewDataBinding.executePendingBindings(ViewDataBinding.java:355)
    at android.databinding.ViewDataBinding$6.run(ViewDataBinding.java:172)
    at android.databinding.ViewDataBinding$5.onViewAttachedToWindow(ViewDataBinding.java:142)
    at android.view.View.dispatchAttachedToWindow(View.java:14531)
    at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:2836)
    at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:2843)
    at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:2843)
    at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:2843)
    at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:2843)
    at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:2843)
    at android.view.ViewGroup.dispatchAttachedToWindow(ViewGroup.java:2843)
    at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:1364)
    at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1107)
    at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:6013)
    at android.view.Choreographer$CallbackRecord.run(Choreographer.java:858)
    at android.view.Choreographer.doCallbacks(Choreographer.java:670)
    at android.view.Choreographer.doFrame(Choreographer.java:606)
[...]

PR #115, opened 2017-02-06
add languageId default value

chiiia12:
## Issue
- None

## Overview (Required)
- AfterParty's language text is empty, I add default value to 'languageId' in SessionViewModel.

## Links
-

## Screenshot
Before | After
:--: | :--:
device-2017-02-06-205946 |device-2017-02-06-205835

PR #118, opened 2017-02-06
fix session lang is null

futabooo:
## Issue
- #117

## Overview (Required)
- fix Resources$NotFoundException when session lang is null

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/util/StringUtil.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/util/StringUtil.java
@@ -1,0 +1,31 @@
+package io.github.droidkaigi.confsched2017.util;
+
+public class StringUtil {
+
+    public static boolean isHalfWidthCharacters(String s) {
+        int[] codePointArray = toCodePointArray(s);// サロゲートペアに対応するためにコードポイントを用いる
+        for (int i = 0; i < codePointArray.length; i++) {
+            int c = codePointArray[i];
+            if ((c <= '\u007e')) {
+                // 英数字
+                continue;
+            } else {
+                // それ以外
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private static int[] toCodePointArray(String str) {
+        int len = str.length();
+        int[] acp = new int[str.codePointCount(0, len)];
+        int j = 0;
+
+        for (int i = 0, cp; i < len; i += Character.charCount(cp)) {
+            cp = str.codePointAt(i);
+            acp[j++] = cp;
+        }
+        return acp;
+    }
+}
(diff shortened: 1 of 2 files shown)


konifar:
👀

app/src/main/java/io/github/droidkaigi/confsched2017/util/StringUtil.java
```diff
@@ -0,0 +1,31 @@
+package io.github.droidkaigi.confsched2017.util;
+
+public class StringUtil {
+
+    public static boolean isHalfWidthCharacters(String s) {
```
konifar: I think it's too much to check the characters 🤔 
If `lang` is null, it should not be bind `lang` to the view and it's better to hide it like this.
https://github.com/DroidKaigi/conference-app-2017/pull/115#discussion_r99573201

What do you think?

konifar:
Thanks for contribution! Commented 😃

app/src/main/java/io/github/droidkaigi/confsched2017/util/StringUtil.java
futabooo: Thank you for review.
I hustle too much 😆 
you are right!

konifar: Anyway, your code is interesting!! 👍 
I learned the logic from your code 📝

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
```diff
@@ -83,6 +84,12 @@ private void setSession(@NonNull Session session) {
 
         if (session.lang != null) {
             this.languageResId = decideLanguageResId(session.lang.toUpperCase());
+        } else {
+            boolean englishTitle = StringUtil.isHalfWidthCharacters(session.title);
+            if (englishTitle) {
+                this.languageResId = decideLanguageResId(Session.LANG_EN_ID);
+            }
+            this.languageResId = decideLanguageResId(Session.LANG_JA_ID);
```
konifar: Does this need `else` ?

konifar:
Sorry, I'll close this PR 🙇 
Thank you for your contribution!!
I would be happy if you will send another PR 🙏

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
futabooo: oh,, right

futabooo:
yes, i will 💪

PR #126, opened 2017-02-06
move TopicColor to viewmodel

YusukeIwaki:
## Issue
-

## Overview (Required)

`TopicColor` contains styles, colors, and so on. 
I think it should not be a model, but a view-model. In fact `TopicColor` is used only by some view-models.

This PR just moves `TopicColor` into the `viewmodel` package :)

## Links
-

## Screenshot

Nothing changed, because of just a refactoring.


konifar:
You are right! It's better.
Thanks for your contribution!!
LGTM :100:

YusukeIwaki:
Thank you for your reviewing so fast :)

PR #122, opened 2017-02-06
Add contributor page (no view yet)

shiraji:
## Issue

#78

## Overview (Required)
Add remote/local logic to get contributors.

For displaying contributors, I will let @srym or others do it. (I will do it if no one does this)
https://github.com/DroidKaigi/conference-app-2017/compare/master...shiraji:78_add_contributor_page?expand=1#diff-4cbabe00489cf45da36d6482d4379adbR77

## Links
N/A

## Screenshot
N/A

Issue #78, opened 2017-02-03
Create Contributors page


konifar:
Assigned to @shiraji

## Overview (Required)
- For great contributors, I want to create Contributors page.
- It would be good to show many contributors in one page.
- GridView is better than ListView?

## Links
- https://github.com/konifar/droidkaigi2016/blob/master/app/src/main/java/io/github/droidkaigi/confsched/activity/ContributorsActivity.java#L72

shiraji:
do you have a concreate design for this page?

konifar:
@shiraji Yes, but it's only in my mind. I'll put the paper design here.

shiraji:
Thanks!

konifar:
@shiraji 
<img src="https://cloud.githubusercontent.com/assets/1269214/22631465/43f76070-ec51-11e6-9c4b-e7cf341944e3.png" width="300" />

- Column 3 grid. Grid is better than List to show many contributors in one page.
- name TextView 14sp, lines="2". You can use `@style/TextBody`
- avator ImageView 60dp.
- Base margin 16dp. You can use `@dimen/space_16dp`

srym:
Is @shiraji going to take this ticket?

shiraji:
Sure but I might be too slow to complete this issue. You can help me out.

Here is what I think we need to do for this issue.

# TODO

- [ ] Create ContributorsRemoteDataSource [sample](https://github.com/konifar/droidkaigi2016/blob/master/app/src/main/java/io/github/droidkaigi/confsched/api/DroidKaigiClient.java#L88)
- [ ] Create ContributorsLocalDataSource using Orma
- [ ] Create ContributorsRepository having above 2 classes (check SessionsRepository)
- [ ] Modify ContributorsViewModel to fetch contributors data
- [ ] Add layout using GridView to ContributorsFragment
- [ ] Add typo for later contributors!

konifar:
>  Add typo for later contributors!

😄

srym:
@shiraji Oh I just want to clarify who is do this task, otherwise I would have made duplicate contribution!
I'd love to help you. What do you think is the best way to collaborate each other? Should I make PR to your forked repository?

shiraji:
Sure. I haven't done anything, yet, thou.

shiraji:
Here is the branch https://github.com/shiraji/conference-app-2017/tree/add_contributor_page

shiraji:
[...]

PR #127, opened 2017-02-06
Convert to ContributorViewModel

shiraji:
## Issue
#78 

## Overview (Required)

* Create `ContributorViewModel` for handling single grid.
* Use `List<ContributorViewModel>`
* Remove callback from `ContributorsViewModel`

## Links
https://github.com/DroidKaigi/conference-app-2017/pull/122#discussion_r99580590

## Screenshot
N/A

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/ContributorViewModel.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/ContributorViewModel.java
@@ -1,0 +1,62 @@
+package io.github.droidkaigi.confsched2017.viewmodel;
+
+import android.databinding.BaseObservable;
+
+import io.github.droidkaigi.confsched2017.model.Contributor;
+
+public class ContributorViewModel extends BaseObservable implements ViewModel {
+
+    private Callback callback;
+
+    private Contributor contributor;
+
+    private String name;
+
+    private String avatarUrl;
+
+    private String htmlUrl;
+
+    private int contributions;
+
+    public ContributorViewModel(Contributor contributor) {
+        this.contributor = contributor;
+        this.avatarUrl = contributor.avatarUrl;
+        this.name = contributor.name;
+        this.htmlUrl = contributor.htmlUrl;
+        this.contributions = contributor.contributions;
+    }
+
+    public void setCallback(Callback callback) {
+        this.callback = callback;
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/ContributorViewModel.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public void destroy() {
+        this.callback = null;
+    }
+
+    public interface Callback {
+
+        void onClickContributor();
```
konifar: This method need `String url` param, right?

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/ContributorViewModel.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private String name;
+
+    private String avatarUrl;
+
+    private String htmlUrl;
+
+    private int contributions;
+
+    public ContributorViewModel(Contributor contributor) {
```
konifar: Please add `@NonNull` parameter 🙇

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/ContributorViewModel.java
shiraji: Is it? Maybe I misunderstand what this ViewModel does...I thougtht some layout call 

```xml
<view  
    onClick=@{viewModel.onClickContributor}> 
```

konifar: No, it's a little different. You have to implement like this.

```java
public void onClickContributor(@SuppressWarning("unused") View view) {
    if (callback != null) {
         callback.onClickContributor(htmlUrl);
    }
}
```

And you can bind this method as you commented.

shiraji: I think it time to go to bed then...sorry about that. You are right.

konifar: No problem! Thanks for contribution! Good sleep 😉

konifar:
LGTM!

PR #140, opened 2017-02-07
save APK into CircleCI's artifact

YusukeIwaki:
## Issue
-

## Overview (Required)

By saving the APK in Circle CI's artifact, we can download and try it easily just with a mobile phone (without PC for build), just by browsing the file like below:

slice

With this feature, we can try the APK even during commute in the train :)
It would be helpful for some people in reviewing and discussing the Pull Requests.

## Links
-

## Screenshot

Nothing changed.


konifar:
Thank you! 👍 LGTM! 💯

PR #141, opened 2017-02-07
Install DeployGate SDK

KeithYokoma:
- This enables us to collect crash logs in DeployGate console

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -203,6 +203,7 @@ dependencies {
     compile 'uk.co.chrisjenx:calligraphy:2.2.0'
 
     // Android Utility
+    compile 'com.deploygate:sdk:3.1.1'
     compile 'com.jakewharton.timber:timber:4.5.1'
     debugCompile 'com.facebook.stetho:stetho:1.4.2'
     debugCompile 'com.facebook.stetho:stetho-okhttp3:1.4.2'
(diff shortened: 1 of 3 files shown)


konifar:
Question 🙋 
I think FirebaseCrash is enough to collect crash logs 🤔 
Is there any other reason to introduce this sdk?

KeithYokoma:
FirebaseCrash is for a production I think.
And if some crash happens, DeployGate immediately notify us via webhook so we can see the report in Slack 😃 .

konifar:
Make sense. Okay. Let's try to taste it 🍸

KeithYokoma:
I've heard that they will soon make some sweet update on the SDK to send a feedback within the app 🎉 . Let's see.

jmatsu:
@KeithYokoma 
No need to add the dependency to only debug?

KeithYokoma:
@jmatsu No concern for the dependency configuration because DeployGate can collect crash logs in a release build!

jmatsu:
Sure! Let's take a taste of it. :wine_glass:

hotchemi:
well, we have a crash report mechanism for debug build as well? I've gotten emails.

konifar:
@hotchemi You mean the crash report for debug build may be noisy?

hotchemi:
@konifar No no I mean we're already collecting crash reports for debug build.
@KeithYokoma Your aim is to integrating Slack and crash report?

KeithYokoma:
@hotchemi Yeah that's my primary intention. In addition to this, we can collect logcat from DeployGate web console remotely in real time if we have permission, so that might help debugging as well.

PR #142, opened 2017-02-07
Change return value Maybe to Completable in SessionDetailViewModel

amay077:
## Issue
- N/A

## Overview (Required)

Changed ``SessionDetailViewModel.findSession()``'s return value ``Maybe<Session>`` to ``Completable``.
Because return value ``session`` does not used in ``SessionDetailFragment``.  ``Completable`` is fit to this case.

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
@@ -88,7 +88,7 @@ public class SessionDetailViewModel extends BaseObservable implements ViewModel 
         this.languageResId = session.lang != null ? decideLanguageResId(session.lang.toUpperCase()) : R.string.lang_en;
     }
 
-    public Completable findSession(int sessionId) {
+    public Completable loadSession(int sessionId) {
         final String languageId = Locale.getDefault().getLanguage().toLowerCase();
         return  sessionsRepository.find(sessionId, languageId)
                 .flatMapCompletable(session -> {
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
@@ -22,7 +22,7 @@ import io.github.droidkaigi.confsched2017.repository.sessions.SessionsRepository
 import io.github.droidkaigi.confsched2017.util.AlarmUtil;
 import io.github.droidkaigi.confsched2017.util.DateUtil;
 import io.github.droidkaigi.confsched2017.util.LocaleUtil;
-import io.reactivex.Maybe;
+import io.reactivex.Completable;
 
 public class SessionDetailViewModel extends BaseObservable implements ViewModel {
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
```diff
@@ -88,12 +88,12 @@ private void setSession(@NonNull Session session) {
         this.languageResId = session.lang != null ? decideLanguageResId(session.lang.toUpperCase()) : R.string.lang_en;
     }
 
-    public Maybe<Session> findSession(int sessionId) {
+    public Completable findSession(int sessionId) {
```
jmatsu: For this purpose, this method should be `existsSession` else.
A method whose name-prefix is`find` should return the associated thingy.

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
```diff
@@ -88,12 +88,12 @@ private void setSession(@NonNull Session session) {
...
         final String languageId = Locale.getDefault().getLanguage().toLowerCase();
-        return sessionsRepository.find(sessionId, languageId)
-                .map(session -> {
+        return  sessionsRepository.find(sessionId, languageId)
+                .flatMapCompletable(session -> {
                     setSession(session);
```
konifar: I think this `finsSession` should return `sessions` though it's not used now.
Maybe this `setSession(session)` make confused to us.
How about changing this method name to `findAndSetSession()` or `initSession()`?
Hmm, I didn't decide good name... Sorry 🙇

k-kagurazaka: I suggest that `findSession` be splited into `findSession` (returns Maybe<Session>) and `refreshUIAsync` (returns Completable).
The current `findSession` contains both command and query.

k-kagurazaka: P.S. The name of `refreshUIAsync` is temporary thing.
Separating responsibility is important.

amay077: Would you like the simple name ``Completable loadSession(sessionId)`` ?
``load`` means read session and sets member.
This contains multiple stuffs but I think enough in this case because only called from ``SessionDetailFragment.onCreate()``.

k-kagurazaka: Nice naming :+1:

amay077: Thanks. I changed it. 👍

konifar:
Thanks! LGTM

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
konifar: 😃

Issue #128, opened 2017-02-06
Android 4.1 Sessions screen become white by scroll


takahirom:
## Overview (Required)
- Android 4.1 Sessions screen become white by scroll
- This confirmed 4.1 Android Emulator 

## Screenshot
![4 1](https://cloud.githubusercontent.com/assets/1386930/22655776/902308e0-ecd4-11e6-8f4f-5a5fb82eb3bd.gif)

PR #143, opened 2017-02-07
Applied a workaround to fix wrong layout when running on api 16

jmatsu:
## Issue
- https://github.com/DroidKaigi/conference-app-2017/issues/128

## Overview (Required)
- It seems twoway-view cannot recognize recycle-span correctly.
- I fixed RecyclerView's height with modifying layout definitions, and then confirmed this can be applied as a workaround of this problem.

## Links
- None

## Screenshot
 Before | After
:--: | :--:
 | api16

Just in case, for API 23
bullheadmmb29qjmatsu02072017030504


konifar:
Thanks! LGTM 👍

PR #149, opened 2017-02-07
sessionId is never used in NotificationReceiver

mapyo:
## Issue
none

## Overview (Required)
I found that sessionId is never used in NotificationReceiver.
So, I delete that.

## Links
none

## Screenshot
none

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/receiver/NotificationReceiver.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/receiver/NotificationReceiver.java
@@ -18,15 +18,13 @@ public class NotificationReceiver extends BroadcastReceiver {
 
     private static final String TAG = NotificationReceiver.class.getSimpleName();
 
-    private static final String KEY_SESSION_ID = "session_id";
     private static final String KEY_TITLE = "title";
     private static final String KEY_TEXT = "text";
 
     private static final int NOTIFICATION_ID = 1;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/receiver/NotificationReceiver.java
```diff
@@ -39,7 +37,6 @@ public void onReceive(Context context, Intent intent) {
             Timber.tag(TAG).v("Notification is disabled.");
             return;
         }
-        int sessionId = intent.getIntExtra(KEY_SESSION_ID, 0);
```
konifar: 🙋 
In my opinion, this should be here.
Because this notification will be improved to open each SessionDetailPage when the user tap the notification.
When we will implement to open SessionDetail, this `sessionId` is necessary.

mapyo: Oh..That's right!
I think so too.

My insight was insufficient 🙇

konifar: No problem! Anyway I'm really grad to your contribution! 😃 
But sorry, let me close this PR 🙇 
I'll be so happy if you send another PR.

konifar: Created the issue! Thanks!
https://github.com/DroidKaigi/conference-app-2017/issues/151

PR #152, opened 2017-02-07
Add open link when contoributors item

nshiba:
## Issue
none

## Overview (Required)
There is no action even if you click on Contributor item.
So we decided to open the url passed in the browser.

## Links
- https://github.com/DroidKaigi/conference-app-2017/issues/78
- https://github.com/DroidKaigi/conference-app-2017/pull/138

## Screenshot
none

PR #153, opened 2017-02-07
Add intent helper

nshiba:
## Issue
none

## Overview (Required)
Since there were many places where intents were created using html url, I tried to create a helper class.
how is it?

## Links
https://github.com/DroidKaigi/conference-app-2017/pull/152#issuecomment-277914167

## Screenshot
none

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/InformationFragment.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/InformationFragment.java
@@ -1,5 +1,7 @@
 package io.github.droidkaigi.confsched2017.view.fragment;
 
+import com.annimon.stream.Optional;
+
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
@@ -112,11 +114,7 @@ public class InformationFragment extends BaseFragment implements InformationView
     }
 
     public void showHtmlUrl(String htmlUrl) {
-        Intent intent = IntentHelper.buildActionViewIntent(getContext(), htmlUrl);
-        if (intent == null) {
-            return;
-        }
-
-        startActivity(intent);
+        Optional<Intent> intentOptional = IntentHelper.buildActionViewIntent(getContext(), htmlUrl);
+        intentOptional.ifPresent(this::startActivity);
     }
 }
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/InformationFragment.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/InformationFragment.java
@@ -2,7 +2,6 @@ package io.github.droidkaigi.confsched2017.view.fragment;
 
 import android.content.Context;
 import android.content.Intent;
-import android.net.Uri;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
 import android.view.LayoutInflater;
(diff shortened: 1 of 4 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/IntentHelper.java
```diff
@@ -0,0 +1,28 @@
...
+
+import timber.log.Timber;
+
+public class IntentHelper {
+
+    /**
+     * Builds an intent of type Intent.ACTION_VIEW from the passed htmlUrl.
+     * But, When intent.resolveActivity(context.getPackageManager()) == null is true, it return null.
+     */
+    public static Intent buildActionViewIntent(@NonNull Context context, @NonNull String htmlUrl) {
```
konifar: Can you add `@Nullable`?
Or how about changing return value to `Optional`?

nshiba: Oh... thanks! 
I trying return `Optional<Intent>`.

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/InformationFragment.java
```diff
@@ -91,42 +88,35 @@ public void showDevInfoPage() {
...
 
     @Override
     public void showYouTube() {
-        Intent intent = new Intent(Intent.ACTION_VIEW, Uri.parse("https://www.youtube.com/droidkaigi"));
-        if (intent.resolveActivity(getContext().getPackageManager()) == null)
+        showHtmlUrl("https://www.youtube.com/droidkaigi");
+    }
+
+    public void showHtmlUrl(String htmlUrl) {
+        Intent intent = IntentHelper.buildActionViewIntent(getContext(), htmlUrl);
```
jmatsu: Just in case, could you please add url format check like `URLUtil.isNetworkUrl`?

app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/IntentHelper.java
jmatsu: @nshiba 
Just in case, could you please add url format check like `URLUtil.isNetworkUrl`?

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/InformationFragment.java
nshiba: 👍

app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/IntentHelper.java
kgmyshin: (or parameter type change to 'Uri'.)

nshiba: @konifar 
Optional requires API 24 or more.
So I add `@Nullable`.

Is there any other good way?

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/InformationFragment.java
kgmyshin: htmlUrl should do null-check or add @NonNull to param.
Because the second param of buildActionViewIntent declare @NonNull

app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/IntentHelper.java
konifar: Oh, really? You couldn't use `com.annimon.stream.Optional`?

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/InformationFragment.java
nshiba: Good point.
But, I think that null check is the role of Helper class, so I will try to null check in helper class with annotation removed

app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/IntentHelper.java
nshiba: Sorry... I didn't not notice.
I use `java.util.Optional`.
thanks!

nshiba:
fixed

konifar:
LGTM! 👍 Thanks!

PR #154, opened 2017-02-07
Remove unnecessary thread control.

sys1yagi:
## Issue

None

## Overview (Required)

Remove unnecessary thread control.
Since updateAllAsync() in SessionsLocalDataSource has no subscriber, I think that obserbeOn() is unnecessary.

## Links

None

## Screenshot

None


konifar:
Good catch! Thank you so much!

konifar:
Thanks!

Issue #139, opened 2017-02-07
Apply constraint layout to view_session_cell.xml


konifar:
## Overview (Required)
Currently, this app doesn't use ConstraintLayout, but there is no reason.
I want to try to use it 😃 

## Links
- https://developer.android.com/reference/android/support/constraint/ConstraintLayout.html

chibatching:
Hi, I can't find `view_session.xml` in the project. Does it means `view_session_cell.xml`?

konifar:
Ah, sorry, right. `view_session_cell.xml` is correct!

chibatching:
Thanks! I'll try it!

PR #157, opened 2017-02-07
Apply constraint layout

chibatching:
## Issue
- Apply constraint layout to view_session_cell.xml #139

## Overview (Required)
- Introduce constraint layout
- Apply constraint layout to view_session_cell.xml
- Update licenses.yml
- Re-generate licenses.html

## Links
- https://developer.android.com/reference/android/support/constraint/ConstraintLayout.html

## Screenshot
Before | After
:--: | :--:
 | 
 |

### DIFF
--- a/app/src/main/assets/licenses.html
+++ b/app/src/main/assets/licenses.html
@@ -540,6 +540,252 @@
     </div>
     <div class="library">
       <!-- https://opensource.org/licenses/Apache-2.0 -->
+      <h1 class="title">DeployGate SDK</h1>
+      <p class="notice">Copyright &copy; DeployGate, tnj, henteko. All rights reserved.</p>
+      <p><a href="https://github.com/DeployGate/deploygate-android-sdk">https://github.com/DeployGate/deploygate-android-sdk</a></p>
+      <div class="license">
+        <h2>
+          Apache License
+          <br/>
+          Version 2.0, January 2004
+          <br/>
+          http://www.apache.org/licenses/
+        </h2>
+        <p>TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION</p>
+        <h2>1. Definitions.</h2>
+        <p>
+          "License" shall mean the terms and conditions for use, reproduction,
+          and distribution as defined by Sections 1 through 9 of this document.
+        </p>
+        <p>
+          "Licensor" shall mean the copyright owner or entity authorized by
+          the copyright owner that is granting the License.
+        </p>
+        <p>
+          "Legal Entity" shall mean the union of the acting entity and all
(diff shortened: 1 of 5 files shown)


konifar:
Cool!! 🆒

chibatching:
I found little change, please see second screenshot diff.

Before, between category border and parent have little space. But I can't confirm why these are occurred.
Parent relative layout had 6dp or 4dp padding and category border had -6dp and -4dp margin.

If these space are intended, I'll add extra space between category boarder and parent layout.

konifar:
CI failed 🤔 

```
> Failed to install the following SDK components:
  [Solver for ConstraintLayout 1.0.0-beta4, ConstraintLayout for Android 1.0.0-beta4]
  Please install the missing components using the SDK manager in Android Studio.
```

app/src/main/assets/licenses.html
```diff
@@ -540,6 +540,252 @@ <h2 class="inline">9. Accepting Warranty or Additional Liability.</h2>
     </div>
     <div class="library">
       <!-- https://opensource.org/licenses/Apache-2.0 -->
+      <h1 class="title">DeployGate SDK</h1>
+      <p class="notice">Copyright &copy; DeployGate, tnj, henteko. All rights reserved.</p>
```
chibatching: In older PR, deploygate was added to license.yml but generateLicensePage task was not executed.

chibatching:
Oops, constraint layout cannot be installed using normal SDK manager (only in Android Studio).
We should download directly from google repository. I'll edit circle.yml, please wait.

### Link
- https://github.com/chibatching/docker-android-sdk/blob/master/Dockerfile#L31-L34

app/src/main/res/layout/view_session_cell.xml
```diff
@@ -32,22 +26,24 @@
 
         <View
                 android:id="@+id/categoryBorder"
-                android:layout_width="match_parent"
+                android:layout_width="0dp"
                 android:layout_height="3dp"
-                android:layout_marginEnd="-6dp"
-                android:layout_marginLeft="-6dp"
-                android:layout_marginRight="-6dp"
-                android:layout_marginStart="-6dp"
```
chibatching: ConstraintLayout doesn't support negative margin. We should add positive margin to other views.

chibatching:
umm

> Your build has exceeded the memory limit of 4G on 1 container. The results of this build are likely invalid. We have taken a snapshot of the memory usage at the time, which you can find in a build artifact named `memory-usage.txt`. The RSS column in this file shows the amount of memory used by each process, measured in kilobytes.

chibatching:
Fix CI failure!

konifar:
Thanks! I'm so excited to use ConstraintLayout! 💪

PR #168, opened 2017-02-07
Group notification

kobakei:
## Issue
-

## Overview (Required)
- Show group notification
  - e.g. If user check 4 sessions which start at 3/9 10:40, 4 notifications will be shown but they should be summarized like Gmail app.
- Set unique ID (=session ID) to PendingIntent and Notification not to override other PendingIntent and Notification objects.

## Links
-

## Screenshot
Before | After
:--: | :--:
 | device-2017-02-07-224530 device-2017-02-07-224542

The second screen shot is captured when user expand group notification.


konifar:
:eyes:

konifar:
LGTM! Thank you! 👏

Issue #167, opened 2017-02-07
Add loading to contributors page


konifar:
## Overview (Required)
Currently there is no loading view.

See `fragment_sessions.xml`.

```xml
        <FrameLayout
                android:layout_width="match_parent"
                android:layout_height="match_parent"
                android:background="@color/white"
                android:visibility="@{viewModel.loadingVisibility}"
                tools:visibility="gone">

            <ProgressBar
                    android:layout_width="wrap_content"
                    android:layout_height="wrap_content"
                    android:layout_gravity="center"
                    />

        </FrameLayout>
```

PR #171, opened 2017-02-07
Added loading to contributors page

t-kashima:
## Issue

- https://github.com/DroidKaigi/conference-app-2017/issues/167

## Overview (Required)

- Show progress bar on loading contributors data

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/ContributorsViewModel.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/ContributorsViewModel.java
@@ -14,7 +14,6 @@ import java.util.List;
 import javax.inject.Inject;
 
 import io.github.droidkaigi.confsched2017.BR;
-import io.github.droidkaigi.confsched2017.model.Contributor;
 import io.github.droidkaigi.confsched2017.repository.contributors.ContributorsRepository;
 import io.reactivex.Single;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/ContributorsViewModel.java
```diff
@@ -50,6 +61,15 @@ public void onClickContributor(String htmlUrl) {
         }
     }
 
+    @Bindable
```
konifar: I think it's not good to check by `contributors`. How about adding `setter` and `getter`?
Like this

```java
private int loadingVisibility;

@Bindable
public int getLoadingVisibility() {
    return loadingVisibility;
}

public void setLoadingVisibility(int visibility) {
    this.loadingVisibility = visibility;
    notifyPropertyChanged(BR.loadingVisibility);
}
```

```
       return contributorsRepository.findAll().map(contributors -> {
            this.contributors = contributors;
            setLoadingVisibility(View.VISIBLE);
```

t-kashima: Thank you for the review.😆 
That's good! I add a loadingVisibility variable.

t-kashima:
I added a variable to switch the visibility😀

konifar:
Good work! LGTM! Thank you for contribution! 👏

PR #121, opened 2017-02-06
Remove ?per_page=100

shiraji:
## Issue

#78 

## Overview (Required)
Remove per_page options from API request.

## Links
https://github.com/DroidKaigi/conference-app-2017/pull/101#discussion_r99518814

## Screenshot
N/A

PR #173, opened 2017-02-07
Add per_page query since github api only returns 30 contributors by default

shiraji:
## Issue
N/A

## Overview (Required)
Previously, we delete the query `per_page` #121 
But I believe this option is mandatory to display more than 30 contributors

## Links
N/A

## Screenshot
Before | After
:--: | :--:
 |


konifar:
Oh I didn't know that! Thanks!

konifar:
I'm so surprised to know there are so many contributors! 😃

shiraji:
If we reach to 100, we need paging. 😉

konifar:
Sounds good! Paging feature will be good example.

Issue #176, opened 2017-02-07
Status bar color in session detail page is wrong.


konifar:
## Overview (Required)
In session detail page, the theme for each topic should be set.

```java
activity.setTheme(viewModel.getTopicThemeResId());
```

But the status bar color is default.

<img width="571" alt="screenshot_jpg__1080x1920_" src="https://cloud.githubusercontent.com/assets/1269214/22696473/70d99d38-ed91-11e6-8d27-3fa121fb6529.png">

PR #183, opened 2017-02-07
Change status bar color when contents scrolled in session detail page

ryugoo:
## Issue
- #176

## Overview (Required)
- Use "colorPrimaryDark" matching the primary color on the session detail page

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionDetailFragment.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionDetailFragment.java
@@ -81,14 +81,11 @@ public class SessionDetailFragment extends BaseFragment implements SessionDetail
             activity.setTaskDescription(taskDescription);
 
             // Change status bar scrim color
-            Window window = activity.getWindow();
-            if (window != null) {
-                TypedValue typedValue = new TypedValue();
-                activity.getTheme().resolveAttribute(R.attr.colorPrimaryDark, typedValue, true);
-                int colorPrimaryDark = typedValue.data;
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionDetailFragment.java
```diff
@@ -85,6 +79,17 @@ private void initTheme() {
                     new ActivityManager.TaskDescription(viewModel.getSessionTitle(), null,
                             ContextCompat.getColor(activity, viewModel.getSessionVividColorResId()));
             activity.setTaskDescription(taskDescription);
+
+            // Change status bar scrim color
+            Window window = activity.getWindow();
```
konifar: I think it's not necessary 🤔 
`activity.setTheme(viewModel.getTopicThemeResId());` is enough to change the theme.
Actually, I could change the status color on DroidKaigi2016 app.
https://github.com/konifar/droidkaigi2016/blob/master/app/src/main/java/io/github/droidkaigi/confsched/fragment/SessionDetailFragment.java#L61

konifar:
Thanks for contribution!

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionDetailFragment.java
ryugoo: Thank you for your comment!
Certainly this code was unnecessary. Delete unnecessary code and push it.

konifar:
LGTM!

PR #185, opened 2017-02-07
Remove Unneeded Semicolon in kt

lvla:
## Issue
- none

## Overview (Required)
- Remove Unneeded Semicolon in kt


konifar:
Good catch! ⚾️ LGTM!

PR #158, opened 2017-02-07
Fix layout preview of fragment_settings.xml on Android Studio

gen0083:
## Issue
-`fragment_settings.xml`'s layout preview was not drawn correctly causes constructor of `SettingsSwitchRowView`.

## Overview (Required)
-

## Links
-

## Screenshot
Before | After
:--: | :--:
before | after

PR #186, opened 2017-02-07
Fix layout preview of InfoRowView on Android Studio

ogapants:
## Issue
none

## Overview (Required)
- Ref #158 
- Fix layout preview of InfoRowView on Android Studio
- and more simple

## Links
none 

## Screenshot
Before | After
:--: | :--:
 |


ogapants:
By the way, I considered [such a code](https://gist.github.com/ogapants/94cebe44b7cdde4552fbd50dee10294b) for preview on Editor, but I stopped it because it is too redundant... 🤔

konifar:
LGTM! 💯 Thanks!!

konifar:
> but I stopped it because it is too redundant...

Yep, I think so too.

PR #188, opened 2017-02-07
log to logcat with Timber

sckm:
## Issue
-

## Overview (Required)
log to logcat when logging with Timber(debug build only)

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


konifar:
LGTM! Thanks for contribution!

PR #189, opened 2017-02-07
Empty txt_date can be seen for a moment, so make it invisible.

syarihu:
## Overview (Required)
- Empty txt_date can be seen for a moment, so make it invisible.

## Screenshot
Before | After
:--: | :--:
 |


jmatsu:
@syarihu 
LGTM! Thank you for improvement!
Now waiting for CI, just in case 😄

syarihu:
@jmatsu Thank you for review!

Issue #181, opened 2017-02-07
Show the number of contributors


konifar:
## Overview (Required)
I want to show the number of contributors.
`Contributors (30 people)`
`コントリビュータ (30人)`

<img width="300" alt="screenshot_jpg__1080x1920_" src="https://cloud.githubusercontent.com/assets/1269214/22700966/a336b9e2-ed9f-11e6-92c3-bb0e3f841ef6.png">

PR #193, opened 2017-02-08
Implements show number of contributors

ryugoo:
## Issue
- #181

## Overview (Required)
- Implements show number of contributors
- Remove unnecessary import

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/ContributorsFragment.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/ContributorsFragment.java
@@ -4,12 +4,11 @@ import com.annimon.stream.Optional;
 
 import android.content.Context;
 import android.content.Intent;
-import android.net.Uri;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.GridLayoutManager;
-import android.util.Log;
 import android.view.LayoutInflater;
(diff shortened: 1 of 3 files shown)


app/src/main/res/values-ja/strings.xml
```diff
@@ -35,6 +35,7 @@
     <string name="sponsors">スポンサー</string>
     <!-- Contributors -->
     <string name="contributors">コントリビュータ</string>
+    <string name="people">(%1$d人)</string>
```
konifar: Can you add `contributors_` prefix?

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/ContributorsFragment.java
```diff
@@ -47,11 +46,11 @@
 
     private Adapter adapter;
 
-    public static ContributorsFragment newInstance() {
-        return new ContributorsFragment();
+    public ContributorsFragment() {
     }
 
-    public ContributorsFragment() {
+    public static ContributorsFragment newInstance() {
```
jmatsu: Could you please align the order of static methods and member methods with other classes? Thanks.

ryugoo: OK. I will fix it!

As for the question, is there a code style definition somewhere?
I refer to "codeStyleSettings.xml", but when I use the code format the constructor and static method will be swapped.

konifar:
LGTM!

PR #201, opened 2017-02-08
Fix notification for API level 16

kobakei:
## Issue
-

## Overview (Required)
- Fix notification for API level 16 ~ 23
  - If group notification is enabled, empty notification will be shown.
  - VIBRATE permission is required for older API versions.

## Links
-

## Screenshot
Before | After
:--: | :--:
device-2017-02-08-133022 | device-2017-02-08-132423

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -6,6 +6,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.VIBRATE" />
 
     <application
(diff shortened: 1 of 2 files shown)


jmatsu:
Good catch!

app/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.VIBRATE" />
```
jmatsu: Could you please add `android:maxSdkVersion="18"`?

konifar:
LGTM!
BTW, it seems better to set large icon to the notification 💭

kobakei:
Okay, I will set large icon.

Issue #205, opened 2017-02-08
Unnecessary page shown when little horizontal scrolled


amay077:
## Overview (Required)

In Sessions page,

- Start and stop vertical scrolling above session item cell -> works fine!
- Start and stop "little" horizontal scrolling above session item cell -> Shows session detail page when scroll ended. it's unnecessary. :scream:
- (Start "big" horizontal scrolling above session item cell -> works file!)

## Screenshot
<img src="https://cloud.githubusercontent.com/assets/401369/22725941/bcd352b0-ee13-11e6-84e1-23a2b7111283.gif" width="300" />

PR #206, opened 2017-02-08
Fix unnecessary page shown when little horizontal scrolled

amay077:
## Issue
- #205

## Overview (Required)

This reason is fired clicked event in RecyclerView because position X does not changes between ACTION_DOWN and ACTION_UP.

I send cancelling motion event when horizontal scrolling.

But, I can't resolve when start scrolling with "long tapped" above session cell.
In this case, still session detail page shown when scroll ended.

## Links

This fix is inspired from ``RecyclerView.java``.

- https://android.googlesource.com/platform/frameworks/support/+/refs/heads/master/v7/recyclerview/src/android/support/v7/widget/RecyclerView.java#1927

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
@@ -8,12 +8,14 @@ import android.content.Context;
 import android.graphics.Point;
 import android.graphics.drawable.Drawable;
 import android.os.Bundle;
+import android.os.SystemClock;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v4.content.res.ResourcesCompat;
 import android.support.v7.widget.RecyclerView;
 import android.text.TextUtils;
 import android.view.Display;
+import android.view.GestureDetector;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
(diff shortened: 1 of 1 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
```diff
@@ -156,11 +158,57 @@ private void initView() {
         adapter = new SessionsAdapter(getContext());
         binding.recyclerView.setAdapter(adapter);
 
+        GestureDetector gestureDetector = new GestureDetector(getContext(), new GestureDetector.OnGestureListener() {
+            private boolean ignoreMotionEventTillDown = false;
+
+            @Override
+            public boolean onDown(MotionEvent motionEvent) {
+                ignoreMotionEventTillDown = true;
```
konifar: Once `ignoreMotionEventTillDown` is set true here, this flag will be never changed.
Is it okay?

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
```diff
@@ -156,11 +158,57 @@ private void initView() {
...
+
         binding.root.setOnTouchListener((v, event) -> {
+            boolean isHandle = gestureDetector.onTouchEvent(event);
+
             MotionEvent e = MotionEvent.obtain(event);
             e.setLocation(e.getX() + binding.root.getScrollX(), e.getY() - binding.headerRow.getHeight());
             binding.recyclerView.forceToDispatchTouchEvent(e);
-            return false;
+
+            return isHandle;
```
jmatsu: Why did you use the return value of new listener?
HorizontalScrolView's touch event `must` not be consumed.

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
amay077: ``ignoreMotionEventTillDown`` changes to ``false`` when ``onScroll``.

konifar: Ah, sorry. Got it.

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
amay077: Thanks, you are right.

amay077: Fixed it!

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
```diff
@@ -156,10 +158,56 @@ private void initView() {
         adapter = new SessionsAdapter(getContext());
         binding.recyclerView.setAdapter(adapter);
 
+        GestureDetector gestureDetector = new GestureDetector(getContext(), new GestureDetector.OnGestureListener() {
```
jmatsu: Could you please create new class or change the location to initialize? because it seems `initView` became so big and many methods of GestureDetector are not used. 

In addition, `ignoreMotionEventTillDown` should be changed to `ignoreMotionEventOnScroll`. It's just typo.
Sorry for splitting reviews. 🙇

amay077: OK 👍

amay077:
I refactored it.
I would like to get your opinion(better class name, method name, etc). :wink:

jmatsu:
Perfectly! LGTM!
@amay077 Thank you for your working!

PR #215, opened 2017-02-08
remove unnecessary try catch in SingleOnSubscribe.subscribe() 

yanzm:
## Issue
- None

## Overview (Required)
- remove unnecessary try catch in SingleOnSubscribe.subscribe() and MaybeOnSubscribe.subscribe(), because MaybeCreate and SingleCreate call onError() when Exception throws in subscribe().

```
public final class SingleCreate<T> extends Single<T> {

    final SingleOnSubscribe<T> source;

    public SingleCreate(SingleOnSubscribe<T> source) {
        this.source = source;
    }

    @Override
    protected void subscribeActual(SingleObserver<? super T> s) {
        Emitter<T> parent = new Emitter<T>(s);
        s.onSubscribe(parent);

        try {
            source.subscribe(parent);
        } catch (Throwable ex) {
            Exceptions.throwIfFatal(ex);
            parent.onError(ex);
        }
    }
    ...
}
```


konifar:
Thank you for improvement! Noted 📝

PR #216, opened 2017-02-08
Fix move to correct session's detail from notifications

gen0083:
## Issue
Session A's notification notified, tap it and move to A's detail.
Next, session B's notification notified, tap it and move to A, not B.

## Overview (Required)
- Extra data of PendingIntent used in notification was not updated.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


konifar:
Good catch! LGTM 👍

PR #218, opened 2017-02-08
Fix stetho interceptor

kobakei:
## Issue
-

## Overview (Required)
- When I open stetho, GitHub API fails.
- Stetho doc says to use `addNetworkInterceptor`.

## Links
- http://facebook.github.io/stetho/#integrations

## Screenshot
Before | After
:--: | :--:
 |


konifar:
Good catch! Thank you! 🙏

konifar:
LGTM!

PR #236, opened 2017-02-08
Manage Disposable with lifecycle in SearchFragment

yatatsu:
## Issue
- None

## Overview (Required)
- SearchFragment subscribed in `onCreateView`, but it was never disposed.
- Clear disposables in `onDestroyView`, and dispose all in `onDetach`.
- And I also destroy viewModel in `onDestroy` (Because it looks to be designed so).

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


konifar:
Good catch! LGTM 💯

Issue #219, opened 2017-02-08
After party cell col span should be 6


konifar:
Assigned to @shiraji 

## Overview (Required)
Currently, after party cell is 1.
But it should be 6 like welcome talk.

<img width="300" alt="screenshot_jpg__1080x1920_" src="https://cloud.githubusercontent.com/assets/1269214/22742614/b829ccf2-ee5a-11e6-914f-27c697ee3c22.png">

## Links
- https://github.com/DroidKaigi/conference-app-2017/blob/master/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionViewModel.java#L115

shiraji:
Here is the value from sessions.json

```
{"id":72,"title":"アフターパーティ / AfterParty","date":"2017/03/10","start_time":"18:30","end_time":"20:30","stime":"2017/03/10 18:30:00","etime":"2017/03/10 20:30:00","duration_min":120,"type":"dinner","desc":"アフターパーティではお寿司をはじめ、美味しい料理とドリンクを用意しています。みんなでワイワイ楽しみませんか？\nhttps://droidkaigi.connpass.com/event/43942/ で参加を受け付けています。\n\n楽しんじゃいましょう、気楽に！"}
```

The interesting part is `"type":"dinner"`. So the question is "is data correct"?

If there is type "dinner", then it seems [enum](https://github.com/DroidKaigi/conference-app-2017/blob/master/app/src/main/java/io/github/droidkaigi/confsched2017/model/Session.java#L88) missing value "dinner". Is there any other type missing?

konifar:
> So the question is "is data correct"?

It's correct. We have to add new enum type `DINNER`.

shiraji:
OK. I will work on this issue.

shiraji:
Is this what you want? It's same colspan with welcome speech but not 6.

![2017-02-09 1 43 47](https://cloud.githubusercontent.com/assets/3675458/22747047/3e67fc18-ee69-11e6-9b35-5aac4b0169b9.png)

konifar:
Thanks!
No, I think the colspan should be 6.

shiraji:
Then, can I fix welcome speech's colspan, too?

![2017-02-09 1 48 40](https://cloud.githubusercontent.com/assets/3675458/22747252/edee2fb8-ee69-11e6-9188-72930ec3ab4e.png)

konifar:
[...]

PR #238, opened 2017-02-08
Make after party col span 6

shiraji:
## Issue
#219 

## Overview (Required)
* Add new session type `DINNER`
* After party cell col span should be 6 

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionViewModel.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionViewModel.java
@@ -114,7 +114,7 @@ public class SessionViewModel extends BaseObservable implements ViewModel {
     private int decideColSpan(@NonNull Session session, int roomCount) {
         if (session.isCeremony()) {
             return 3;
-        } else if (session.isBreak()) {
+        } else if (session.isBreak() || session.isDinner()) {
             return roomCount;
         } else {
             return 1;
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/model/Session.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/model/Session.java
@@ -84,7 +84,7 @@ public class Session {
     public String shareUrl;
 
     private enum Type {
-        CEREMONY, SESSION, BREAK;
+        CEREMONY, SESSION, BREAK, DINNER;
 
         boolean matches(String type) {
             return name().equalsIgnoreCase(type);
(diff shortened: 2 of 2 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionViewModel.java
```diff
@@ -117,6 +117,8 @@ private int decideColSpan(@NonNull Session session, int roomCount) {
             return 3;
         } else if (session.isBreak()) {
             return roomCount;
+        } else if (session.isDinner()) {
+            return 6;
```
konifar: This is better 😃 

```java
} else if (session.isBreak() || session.isDinner()) {
    return roomCount;
```

konifar: Sorry, my explanation was lacked 🙇

shiraji: Does `roomCount` always 6? If so, we could change `Type` to the following code.

```java
private enum Type {
        CEREMONY(3), SESSION(1), BREAK(6), DINNER(6);

        int colspan;

        Type(int colspan) {
            this.colspan = colspan;
        }

        @Override
        public String toString() {
            return super.toString().toLowerCase();
        }
    }
```

and then remove `isBreak()` etc methods, and these if/else statements.

konifar: `roomCount` is always 6 on DroidKaigi2017. But it's calculated by sessions data.

And in my opinion, Session.Type should not have UI data like colSpan.
It's better to switch colspan in ViewModel.
What do you think?

shiraji: OK now got it. I think `Session.Type` is weird. It should not be there.

shiraji: never mind. let me sleep tonight. I will think about this one.

By the way, I totally agree with what you wrote. colspan should be handled in ViewModel.

konifar: Thank you! Sorry, coflict occurred. Can you merge master tomorrow? 🙇

shiraji:
done + rebase the commits.

After spent one night, I still cannot figure out the best practice...so I fixed the if/else statements. If I found something later, I will send other PR.

konifar:
Thanks! Yes, I don't think this is the best way too. Anyway, I appreciate your work!
LGTM 👍

Issue #161, opened 2017-02-07
Change session cell background color on the current time


konifar:
## Overview (Required)

<img width="671" alt="screenshot_jpg__1080x1920_" src="https://cloud.githubusercontent.com/assets/1269214/22690211/023d467c-ed78-11e6-9fe7-036dfe9e3154.png">

PR #239, opened 2017-02-08
Highlight ongoing sessions

ytRino:
## Issue
- #161

## Overview (Required)
- Highlight ongoing sessions. It checks every session with `new Date()` but I don't have any good pattern for this and implemented such roughly as-is.
- It seems `Session.Type#toString()` is overriden just for `equals` check.  
If `Session.type` is `Type type`, `Session#isBreak()` will be `type == Type.BREAK`.
idk why `Session.type` is not `Type` (im not familiar with orma, adapting enum brings complexity?) so just make `Type` to have its own equality method using `name().equalsIgnoreCase()`. I will revert this change as you like (eg. json should strictly declare lowercase for `type` key. this change will be breaking).

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/model/Session.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/model/Session.java
@@ -109,6 +109,10 @@ public class Session {
         return type.toString().equals(this.type);
     }
 
+    public boolean isLiveAt(Date when) {
+        return stime.before(when) && etime.after(when);
+    }
+
     @Override
     public boolean equals(Object o) {
         return o instanceof Session && ((Session) o).id == id || super.equals(o);
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/model/Session.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/model/Session.java
@@ -6,7 +6,6 @@ import com.github.gfx.android.orma.annotation.Column;
 import com.github.gfx.android.orma.annotation.PrimaryKey;
 import com.github.gfx.android.orma.annotation.Table;
 
-import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


konifar:
:eyes:

konifar:
> why Session.type is not Type (im not familiar with orma, adapting enum brings complexity?)

As you said, it's better to change `type` to enum since orma supports enum.
Anyway, I'm for your change.

app/src/main/res/drawable-v21/clickable_purple.xml
```diff
@@ -0,0 +1,13 @@
...
+<ripple xmlns:android="http://schemas.android.com/apk/res/android"
+        android:color="@color/black_alpha_12"
+        >
+
+    <item
+            android:id="@android:id/mask"
+            android:drawable="@android:color/white"
+            />
+
+    <item android:drawable="@color/purple_alpha_15" />
```
konifar: I specified this color, but maybe the different color is better.
It's another issue, so no need to change it on this PR! 👍

konifar:
LGTM! Thanks! 💯

app/src/main/java/io/github/droidkaigi/confsched2017/model/Session.java
```diff
@@ -87,26 +86,25 @@
     private enum Type {
         CEREMONY, SESSION, BREAK;
 
-        @Override
-        public String toString() {
-            return super.toString().toLowerCase();
+        boolean matches(String type) {
+            return name().equalsIgnoreCase(type);
```
jmatsu: `equalsIgnoreCase` depends on device's Locale, so can you use `name().toLowerCase(Locale.US).equals(type.toLowerCase(Locale.US))`?

jmatsu: Sorry, my review is late. Please don't mind.

konifar: Oh, sorry... Your review makes sense! Let's fix it on another PR 💪

ytRino: wow 😮 interesting.
I didn't know that and take a quick search, then javadoc and some blog posts are pointing 
that `equalsIgnoreCase()` is locale *in*dependent.
https://developer.android.com/reference/java/lang/String.html#equalsIgnoreCase(java.lang.String)
http://yuki312.blogspot.jp/2013/05/androidtolowercasetouppercase.html
I'm not tested myself but if you have some clarification, plz let me know. thx!

PR #241, opened 2017-02-08
Return to session tab when back pressed on MainActivity

chibatching:
## Issue
- none

## Overview (Required)
- Return to session tab when back pressed on other tabs on MainActivity
  - ref: Google+ app, Twitter app

## Links
- none

## Screenshot
Before | After
:--: | :--:
 |


konifar:
Thanks for improvement!
Looks good 👍

konifar:
LGTM! Thanks 👏

Issue #243, opened 2017-02-09
Session language in detail page is always English


chibatching:
Assigned to @chibatching 

## Overview (Required)
- Session language in detail page is always English even if it is Japanese session.

## Links
- none

## Screenshot
<img src="https://cloud.githubusercontent.com/assets/7804631/22764434/0d2b4070-eeae-11e6-81e5-87ceceb5f384.png" width="300" />

PR #244, opened 2017-02-09
Fix session language decision in SessionDetailViewModel

chibatching:
## Issue
- Session language in detail page is always English #243

## Overview (Required)
- Change session id letter case from upper to small, because `LocaleUtil` constant value is lower case.

## Links
- https://github.com/DroidKaigi/conference-app-2017/blob/master/app/src/main/java/io/github/droidkaigi/confsched2017/util/LocaleUtil.java

## Screenshot
Before | After
:--: | :--:
 |


konifar:
:eyes:

konifar:
Good catch! LGTM! 💯  Thanks!

Issue #246, opened 2017-02-09
Detect memory leaks of com.tomoima.debot.Debot in SessionDetailActivity


amay077:
Assigned to @tomoima525

## Overview (Required)

I'm testing app installed by DeployGate ＃123.
LeakCanary caught memory leaks in SessionDetailActivity.

### Steps for REPRO

1. Launch app and shown sessions page
2. Tap to any session and open session detail page
3. Back to session page with tap to ← button
4. Wait a some minutes
5. LeakCanary reports leaks

I think this issue is related to [PR 207](https://github.com/DroidKaigi/conference-app-2017/pull/207).
@tomoima525, Could you help us?

## Links
- https://github.com/DroidKaigi/conference-app-2017/pull/207

## Screenshot
<img src="https://cloud.githubusercontent.com/assets/401369/22770679/8eb87f78-eed5-11e6-9824-a0b09db2f450.gif" width="300" />

## Dump

* [dump.txt](https://github.com/DroidKaigi/conference-app-2017/files/762691/dump.txt)

tomoima525:
got it, let me take a look

konifar:
@tomoima525 I assigned you, but PR has been sent...! Too fast...
https://github.com/DroidKaigi/conference-app-2017/pull/248

ytRino:
oops sorry i didnt noticed for comment while working on this :bow:

konifar:
@ytRino Never mind! Thank you for super quick fix! ⚡️

tomoima525:
Oh my god, you are awesome !! 👍  @ytRino

PR #248, opened 2017-02-09
Use application context for sensor

ytRino:
## Issue
- #246 

## Overview (Required)
- Leakcanary reports leaking Activity in Debot. It seems SensorManager might being hold Activity if created with `activityContext.getSystemService()` 
 https://github.com/tomoima525/debot/blob/master/debot/src/main/java/com/tomoima/debot/Debot.java#L57 
I think it can be solved with some ways.
1. Debot uses `context.getApplicationContext().getSystemService()`
2. Passing `getApplicationContext()` by ourselves.
3. Move calling `debot#allowShake()` to `onResume()`
4. AOSP fix this.

I did quick fix with 2 and not detected leak currently. But not much tested, would you check this?


konifar:
Looks good.
@tomoima525 What do you think?

tomoima525:
lgtm, let's ship it 🚀  Thank you for the quick fix!

konifar:
🚀

PR #252, opened 2017-02-09
Fix label in README.md

noboru-i:
## Issue
- None

## Overview (Required)
- I think, I found unnecessary `)` in https://github.com/DroidKaigi/conference-app-2017#contributing . So I delete that.  
If that is not mistake, so sorry.

## Links
- None

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 
 # Contributing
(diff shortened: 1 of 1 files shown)


konifar:
Good catch! Thanks! LGTM 💯

PR #253, opened 2017-02-09
Add touch feedback to information icons with ripple

ogapants:
## Issue
none

## Overview (Required)
According to the material design guidelines, view that can be clicked need touch feedback with ripple effect.
(In pre-lollipop, this setting is ignored.)

## Links
https://material.io/guidelines/motion/choreography.html#choreography-radial-reaction

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/res/layout/fragment_information.xml
+++ b/app/src/main/res/layout/fragment_information.xml
@@ -67,6 +67,7 @@
                             android:id="@+id/img_twitter"
                             android:layout_width="@dimen/info_social_icon_size"
                             android:layout_height="@dimen/info_social_icon_size"
+                            android:background="?attr/selectableItemBackgroundBorderless"
                             android:onClick="@{viewModel::onClickTwitter}"
                             app:srcCompat="@drawable/ic_twitter_36_vector"
                             />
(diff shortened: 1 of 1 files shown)


konifar:
:eyes:

konifar:
LGTM! Thanks 👍

app/src/main/res/layout/fragment_information.xml
```diff
@@ -77,6 +78,7 @@
                             android:layout_height="@dimen/info_social_icon_size"
                             android:layout_marginLeft="@dimen/info_social_icon_margin"
                             android:layout_marginStart="@dimen/info_social_icon_margin"
+                            android:background="?attr/selectableItemBackgroundBorderless"
```
konifar: BTW, I think it's better to extract some attributes to styles.xml 💭

ogapants: Like this? 🤔 
```xml
    <style name="ClickCover.Borderless" parent="ClickCover">
        <item name="android:background">?attr/selectableItemBackgroundBorderless</item>
    </style>
```

PR #286, opened 2017-02-10
change argument type(ActivityComponent#inject)

yyYank:
## Issue

I think ActivityComponent inject method is too many.
My proposal is change argument type. 
<del>My proposal is use type parameter. Generic type parameter solves this issue.</del>

## Overview (Required)

ActivityComponent inteface seems simply inject Activity instance.
But it's provides overload  methods each subclass :

```java
void inject(XXXActivity activity);
```

This approach cause increase method when Activity added.

I thought BaseActivity is boundary type for inject method in droidkaigi app.
So inject method definition is :

<del>
```java
T extends BaseActivity
```
</del>

```
void inject(BaseActivity activity);
```

Honestly, I want to use generic type parameter,but Dagger2 seems not allow it.

## Links

* Effective Java(Item 28: Use bounded wildcards to increase API flexibility. )
* [Dagger 2 Generic Type class inject error](http://stackoverflow.com/questions/32498912/dagger-2-generic-type-class-inject-error)

## Screenshot

None.

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/di/ActivityComponent.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/di/ActivityComponent.java
@@ -10,7 +10,7 @@ import io.github.droidkaigi.confsched2017.view.activity.*;
 @Subcomponent(modules = ActivityModule.class)
 public interface ActivityComponent{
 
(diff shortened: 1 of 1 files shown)


konifar:
Thank you for good suggestion 😃 
It looks better. LGTM!

PR #290, opened 2017-02-11
fix wrong label

noboru-i:
## Issue
- None

## Overview (Required)
- Before: `ContributorsActivity` has "Sponsors" as label.  
Seems that is wrong.  
I think, "Contributors" is ok.

## Links
- None

## Screenshot
None

By the way, we really need `android:label` property for not `LAUNCHER` Activity...?
https://developer.android.com/guide/topics/manifest/activity-element.html?hl=ja#label

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -52,7 +52,7 @@
                 android:name=".view.activity.ContributorsActivity"
                 android:configChanges="keyboard|keyboardHidden|screenLayout|screenSize|orientation"
                 android:exported="false"
-                android:label="@string/sponsors"
+                android:label="@string/contributors"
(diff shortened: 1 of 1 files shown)


konifar:
Good catch! ⚾️ 
LGTM! 💯

konifar:
Thanks for contribution!

noboru-i:
Thanks!

Sorry for small changes... :sweat_smile:

konifar:
> By the way, we really need android:label property for not LAUNCHER Activity...?

Hmm, I'm not sure. But it looks unnecessary. Let me investigate more.

konifar:
> Sorry for small changes... 😅

Actually I didn't know this issue. So this PR is really worth for us 👍

PR #298, opened 2017-02-11
Remove unused SessionFeedbackViewModel injection.

k-kagurazaka:
## Issue
None

## Overview (Required)
- `SessionFeedbackViewModel` is not used in `SessionFeedbackActivity` so I have removed it's injection.


konifar:
Thanks! Maybe this viewmodel will be used when we will implement feedback page.
But it's okay to remove it now.

konifar:
LGTM!

k-kagurazaka:
I think that feedback page will be implemented in `SessionFeedbackFragment` and the activity simply has this fragment.
Do I misunderstand something?

PR #301, opened 2017-02-11
Fixed the output of the log message by Timber's recommended way.

horie1024:
## Issue
- None

## Overview (Required)
- `BinaryOperationInTimber` occurred when executing lint.
  - This is caused by a [lint rule embedded in Timber](https://github.com/JakeWharton/timber#lint).
- Timber internally uses `String.format` to format log messages.
- Therefore, I changed the format of the log message to the way Timber recommends.

## Links
- None

## Screenshot
Before | After
:--: | :--:
 |


konifar:
LGTM! Thank you so much! 👍

PR #306, opened 2017-02-11
Unify method of calling Activity to `createIntent`

ogapants:
## Issue
none

## Overview (Required)
Two kinds of methods of calling Activity exist in the current project.
First, `start` method in ContributorsActivity, MainActivity, SearchActivity and SponsorsActivity.
Second, `createIntent` method in SessionDetailActivity and SessionFeedbackActivity.

These should be unified into either.
I think it is better to use `createIntent` because it can use `Intent#addFrag` and `startActivityForResult` later... 🤔 

## Links
none

## Screenshot
none


konifar:
You're right.
As you said, we adopted `createdIntent()` on `SessionDetailActivity` and `SessionFeedbackActivity`, and I think it's better than `start()` method.

I'm for this change. LGTM!

PR #307, opened 2017-02-12
Add assertsutil test

kitwtnb:
## Issue
- none

## Overview (Required)
- Add AssertUtilTest.

## Links
- none

PR #310, opened 2017-02-12
Remove unnecessary test case.

sys1yagi:
## Issue

None

## Overview (Required)

Even if the DateUtil constructor does not throw an exception the original test succeeds.
I introduced ExpectedException.

## Links

None

## Screenshot

None


konifar:
:eyes:

sys1yagi:
I think this test is unnecessary. ref https://github.com/DroidKaigi/conference-app-2017/pull/307#discussion_r100682136

What do you think?

konifar:
@sys1yagi Agree. If you agree, can you remove this test in this PR? 🙇

sys1yagi:
of cause :)

sys1yagi:
done~

konifar:
Thanks! LGTM 👍

PR #311, opened 2017-02-12
Use method references.

sys1yagi:
## Issue

None

## Overview (Required)

Use method references in stream api processing.

## Links

None

## Screenshot

None


konifar:
Good catch! LGTM 💯  Thanks!

Issue #180, opened 2017-02-07
Stop using twoway-view


konifar:
## Overview (Required)
- I introduced twoway-view to implement recyclerview with colspan and rowspan.
- But it has some bugs.
- I want to stop using twoway-view.

## Links
- https://github.com/lucasr/twoway-view/

amay077:
I think so too.

LeakCanary caught "EXCLUDED" leak in MainActivity when exit app.
I suspect it is related to twoway-view.

FYI: attached the [LeakCanary log file](https://github.com/DroidKaigi/conference-app-2017/files/763602/leakcanary_dump.txt)

eyedol:
How about a `ViewPager` with indicators( they will be the room numbers) and a recyclerview with a grid layout manager to display the room's respective sessions.

PR #312, opened 2017-02-12
Ignore click events on session header

tnj:
Prevent session cell click event triggered unexpectedly by tapping on header view.

## Issue
- Nothing related

## Overview (Required)
- Just a quick minor UI fix: tapping on the header view of session list causes click event on the cell behind. I don't think that is intended behavior.
- I just found #180, so this change might be a short life, though.

## Links
- None

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/res/layout/view_sessions_header_cell.xml
+++ b/app/src/main/res/layout/view_sessions_header_cell.xml
@@ -7,4 +7,5 @@
         android:gravity="center"
         android:textColor="@color/black"
         android:textStyle="bold"
+        android:clickable="true"
         />


konifar:
:eyes:

konifar:
Good catch! Thanks!

> I just found #180, so this change might be a short life, though.

Yep, but #180 seems difficult. Maybe it will be a long life 😄

konifar:
LGTM 💯

tnj:
Wow, superfast review 🚀  Thanks!

Issue #280, opened 2017-02-10
Stop using deprecated `updateConfiguration`


konifar:
## Overview (Required)
https://github.com/DroidKaigi/conference-app-2017/blob/master/app/src/main/java/io/github/droidkaigi/confsched2017/util/LocaleUtil.java#L37-L50

`updateConfiguration` has been deprecated in API 25.
https://developer.android.com/reference/android/content/res/Resources.html#updateConfiguration(android.content.res.Configuration, android.util.DisplayMetrics)

So it's better to change the way to apply Locale.

## Links
- https://github.com/DroidKaigi/conference-app-2017/pull/270#discussion_r100486650

PR #315, opened 2017-02-12
[WIP] stop using updateConfiguration

rei-m:
## Issue
- #280 

## Overview (Required)
- SetUp Locale on attachBaseContext. And Pass that to CalligraphyContextWrapper.
- Remove SetUp Context from LocaleUtil.

## Links
- http://stackoverflow.com/questions/40221711/android-context-getresources-updateconfiguration-deprecated

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/BaseActivity.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/BaseActivity.java
@@ -3,6 +3,8 @@ package io.github.droidkaigi.confsched2017.view.activity;
 import com.tomoima.debot.Debot;
 
 import android.content.Context;
+import android.content.res.Configuration;
+import android.os.Build;
 import android.os.Bundle;
 import android.support.annotation.IdRes;
 import android.support.annotation.LayoutRes;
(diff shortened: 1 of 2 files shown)


konifar:
:eyes:

konifar:
Thanks! 👍 
I have a question 🙋 

> SetUp Locale on attachBaseContext. And Pass that to CalligraphyContextWrapper.

The locale is initialized in BaseActivity on this PR.
Is it applied to non-activity class such as Notification?

rei-m:
> The locale is initialized in BaseActivity on this PR.
> Is it applied to non-activity class such as Notification?

No... This Setting is applied only activities. Locale of Notification is SystemLocale.
I searched to apply configuration that created by context.createConfigurationContext, and found this solution only.

app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/BaseActivity.java
```diff
@@ -99,4 +107,28 @@ final void initBackToolbar(Toolbar toolbar) {
             bar.setHomeButtonEnabled(true);
         }
     }
+
+    @SuppressWarnings("deprecation")
+    private Context setUpLocale(Context context, String language) {
```
shiren-wanderer: I believe this method should be kept in LocaleUtil.java because this is for "Locale" update.

shiren-wanderer:
As @rei-m mentioned, this is only for activities but not non-activity stuff.
The main issue here is that createConfigurationContext() just returns context instead of directly updating it. Thus, MainApplication's context only can have the latest locale after app restart.

konifar:
@rei-m Sorry, I'm not sure this is correct fix or not.
Please let me investigate more and wait more 🙇

rei-m:
@shiren-wanderer Thanks 🙇  I'll try to setup at  Application's attachBaseContext.

@konifar OK. I continue to investigation too.

konifar:
Sorry for too late review 🙇 🙇 🙇 
In my investigation, latest Android doesn't want to allow changing locale from app 🤔 
So I want to move this locale change option to debug menu.
If you agree, I'll create another issue.
What do you think?

rei-m:
@konifar Ok!! I agree👌 then please close this PR and create another issue 🙇

PR #319, opened 2017-02-12
[Showcase Feature] Overlay view to show timber logs for debugging

KeithYokoma:
## Overview (Required)
- I am going to talk about "drawing views on overlay" and this is an example implementation of it.
- I added "Debug Overlay View" setting and if you enable it and Timber logs emitted, log messages are displayed on top of any apps.
- The setting is default off. A service responsible to drawing views on overlay will be running during the process is alive and the setting is enabled.
- If you remove DroidKaigi app from a recent task list, the overlay also goes away. This means the service is also stopped.
- Handles configuration changes in the service and update views on overlay.

## Screenshots

 | 

capture


konifar:
:eyes:

konifar:
Interesting 🆒

konifar:
How about adding `Developer menu` title here?

_showcase_feature__overlay_view_to_show_timber_logs_for_debugging_by_keithyokoma_ _pull_request__319_ _droidkaigi_conference-app-2017

app/src/main/java/io/github/droidkaigi/confsched2017/di/OverlayViewComponent.java
```diff
@@ -0,0 +1,14 @@
...
+import dagger.Subcomponent;
+import io.github.droidkaigi.confsched2017.di.scope.OverlayViewScope;
+import io.github.droidkaigi.confsched2017.view.customview.DebugOverlayView;
+
+/**
+ * @author KeithYokoma
+ */
+@OverlayViewScope
+@Subcomponent(modules = {OverlayViewModule.class})
+public interface OverlayViewComponent {
```
konifar: Question 🙋 
To be honest, I don't know how the modules should be divided.
If you have any opinion, please you tell me 🙇

app/src/main/java/io/github/droidkaigi/confsched2017/service/BaseService.java
```diff
@@ -0,0 +1,33 @@
...
+    public ServiceComponent getComponent() {
+        if (serviceComponent == null) {
+            MainApplication mainApplication = (MainApplication) getApplication();
+            serviceComponent = mainApplication.getComponent().plus(new ServiceModule(this));
+        }
+        return serviceComponent;
+    }
+
+    @Retention(RetentionPolicy.SOURCE)
+    @IntDef(value = {Service.START_FLAG_REDELIVERY, Service.START_FLAG_RETRY}, flag = true)
```
konifar: 📝

app/src/main/java/io/github/droidkaigi/confsched2017/service/DebugOverlayService.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Created by KeishinYokomaku on 2017/02/12.
+ */
+
+public class DebugOverlayService extends BaseService {
+    private final BroadcastReceiver configurationChangeReceiver = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            handler.post(() -> {
+                Timber.d("onConfigurationChange");
```
konifar: Is it better to add `.tag(TAG)`?

app/src/main/java/io/github/droidkaigi/confsched2017/util/ContextUtil.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @author KeithYokoma
+ */
+public final class ContextUtil {
+
+    private ContextUtil() {
+        throw new AssertionError("no instance");
+    }
+
+    public static Context getBaseContext(Context context) {
```
konifar: 🙋 
Is this util class necessary? 🤔 I think this util method is used in few places.

konifar: I'm afraid that many util class makes the contributors confused.

app/src/main/java/io/github/droidkaigi/confsched2017/view/customview/DebugOverlayView.java
```diff
@@ -0,0 +1,62 @@
...
+    public DebugOverlayView(Context context, AttributeSet attrs) {
+        this(context, attrs, 0);
+    }
+
+    public DebugOverlayView(Context context, AttributeSet attrs, int defStyleAttr) {
+        super(context, attrs, defStyleAttr);
+        binding = ViewOverlayDebugBinding.inflate(
+                (LayoutInflater) context.getSystemService(Context.LAYOUT_INFLATER_SERVICE), this, true);
+        // XXX
+        ((DebugOverlayService) ContextUtil.getBaseContext(ContextUtil.getBaseContext(context)))
```
konifar: `ContextUtil.getBaseContext(context)` is enough?

app/src/main/java/io/github/droidkaigi/confsched2017/view/customview/DebugOverlayView.java
```diff
@@ -0,0 +1,62 @@
...
+                .getComponent().plus(new OverlayViewModule()).inject(this);
+    }
+
+    @Override
+    protected void onAttachedToWindow() {
+        super.onAttachedToWindow();
+        disposables.add(emitter.listen().subscribe(log -> {
```
konifar: 👍

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SettingsFragment.java
```diff
@@ -95,6 +101,25 @@ public void showLanguagesDialog() {
...
 
+    @Override
+    public void debugOverlayViewEnabled(boolean enabled) {
+        if (isDetached())
+            return;
+        if (enabled && !SettingsUtil.canDrawOverlays(getContext())) {
+            Timber.tag(TAG).d("not allowed to draw views on overlay");
+            binding.debugOverlayViewSwitchRow.setChecked(false);
+            Toast.makeText(getContext(), R.string.settings_debug_overlay_view_toast, Toast.LENGTH_LONG).show();
+            Intent intent = new Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION, Uri.parse("package:" + getContext().getPackageName()));
```
konifar: Can you extract these to `IntentHelper`?

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SettingsViewModel.java
```diff
@@ -71,6 +75,13 @@ public void onCheckedNotificationSetting(boolean isChecked) {
         }
     }
 
+    public void onCheckedDebugOverlayView(boolean isChecked) {
+        defaultPrefs.putShowDebugOverlayView(isChecked);
+        if (callback == null)
+            return;
```
konifar: I like this style. No need to fix it 😃 

```java
if (callback != null) {
    callback.debugOverlayViewEnabled(isChecked);
}
```

app/src/main/res/layout/fragment_settings.xml
```diff
@@ -139,6 +139,21 @@
                     android:visibility="@{viewModel.getShowHeadsUpSettingVisibility}"
                     />
```
konifar: How about to add title?

```java
        <RelativeLayout
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
                     android:padding="@dimen/space_16dp"
                     >
 
                 <TextView
                         style="@style/TextSubheading"
                         android:layout_width="match_parent"
                         android:layout_height="wrap_content"
                         android:text="Developer menu"
                         android:textColor="?attr/colorPrimary"
                         />
 
             </RelativeLayout>
```

app/src/main/res/layout/view_overlay_debug.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+        >
+    <RelativeLayout
```
konifar: Is this `<RelativeLayout>` necessary?

konifar:
Commented! Generally I agree to add this cool show case 👍

app/src/main/java/io/github/droidkaigi/confsched2017/di/AppModule.java
```diff
@@ -52,6 +53,12 @@ public SharedPreferences provideSharedPreferences(Context context) {
     }
 
     @Provides
+    @Singleton
+    public LogEmitter provideLogEmitter() {
+        return new LogEmitter();
```
k-kagurazaka: I think no need explicit provider for `LogEmitter`.
How about use following?
```java
@Singleton
public class LogEmitter {
...
```

app/src/main/java/io/github/droidkaigi/confsched2017/service/helper/OverlayViewManager.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+import io.github.droidkaigi.confsched2017.R;
+import uk.co.chrisjenx.calligraphy.CalligraphyContextWrapper;
+
+/**
+ * @author KeithYokoma
+ */
+@Singleton
```
k-kagurazaka: This `@Singleton` annotation does nothing.
Currently `OverlayViewManager` is provided by `ServiceModule`, so we should add `@ServiceScope` to the provider method if you want to control the lifecycle of this class instance.

Or if `Application` instance can be used for `Context` argument of this class constructor, we can simply remove provider of this class.

app/src/main/java/io/github/droidkaigi/confsched2017/di/OverlayViewComponent.java
k-kagurazaka: @konifar Are you going to say how the "components" should be devided?
If so, I have the same question.

k-kagurazaka:
Some commented.
It's a super cool feature:heart_eyes:

app/src/main/java/io/github/droidkaigi/confsched2017/di/OverlayViewComponent.java
konifar: Yes, that's my question.

app/src/main/java/io/github/droidkaigi/confsched2017/service/DebugOverlayService.java
KeithYokoma: Actually Timber does that for us without calling `.tag(TAG)` explicitly so unless some tag customization is needed we don't need to call it.

app/src/main/java/io/github/droidkaigi/confsched2017/view/customview/DebugOverlayView.java
KeithYokoma: It's not enough because `DebugOverlayService` is wrapped with `ContextThemeWrapper` and the wrapper is wrapped `ContextWrapper` again x(

app/src/main/java/io/github/droidkaigi/confsched2017/di/OverlayViewComponent.java
KeithYokoma: I think this is too much for this kind of small feature, but `Service` is too big and I would like to introduce the similar architecture on the overlay views to Fragments.

app/src/main/java/io/github/droidkaigi/confsched2017/util/ContextUtil.java
KeithYokoma: ok, will remove this!

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SettingsFragment.java
KeithYokoma: 👀

app/src/main/java/io/github/droidkaigi/confsched2017/di/AppModule.java
KeithYokoma: 👍 Thanks!

app/src/main/java/io/github/droidkaigi/confsched2017/service/helper/OverlayViewManager.java
KeithYokoma: I forgot removing this sorry!

app/src/main/java/io/github/droidkaigi/confsched2017/service/BaseService.java
KeithYokoma: Actually this annotation is already defined in `Service` class but is hidden!!!!

app/src/main/res/layout/view_overlay_debug.xml
KeithYokoma: removing!

KeithYokoma:
Thanks for feedback 🙇 and fixed issues!

And let me explain about module separation.

Currently this app doesn't have so many services and views in the overlay so `OverlayViewModule` can be merged into `ServiceModule`.
If the app has more views and building some back stack like Fragment then it's time to separate the module. Or having more services as well.

If it's okay to leave it as it is then I will put Javadoc on related classes, otherwise I will merge modules into `ServiceModule`.
What do you thing? @konifar @k-kagurazaka

konifar:
Thanks! I'll check your comments tonight!

k-kagurazaka:
@KeithYokoma 
OK, I just understand your plan.
As you said, `OverlayViewScope` has the different and shorter life time than `ServiceScope`, so separating component is good pattern 👍 

I also agree with `OverlayViewModule`.
Currently it provides no dependencies, however I know the existence of `OverlayViewModule` class indicates the place where do we add providers for `OverlayViewScope`.

konifar:
@KeithYokoma Understood. Thanks! Please go ahead as it is 🏃

konifar:
LGTM! 👍

PR #320, opened 2017-02-12
Change map to doOnSuccess

STAR-ZERO:
## Issue
- None

## Overview (Required)

map operator purpose is to transform items.
In this case, Using doOnSuccess is better.

## Links
- [ReactiveX \- Map operator](http://reactivex.io/documentation/operators/map.html)
- [ReactiveX \- Single](http://reactivex.io/documentation/single.html)

## Screenshot
Before | After
:--: | :--:
 |


konifar:
Thanks for improvement! 

> map operator purpose is to transform items.
> In this case, Using doOnSuccess is better.

You're right. Your code absolutely looks better 👍 
LGTM!

PR #322, opened 2017-02-12
Fix "tools:context" in layout.xml

ogapants:
## Issue
none

## Overview (Required)
I fix wrong "tools:context" in layout.xml.
Also add to layout.xml of Activity which did not have "tools:context".

## Links
none

## Screenshot
There was no change on the Android Studio preview... 🙄

### DIFF
--- a/app/src/main/res/layout/activity_contributors.xml
+++ b/app/src/main/res/layout/activity_contributors.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <layout xmlns:android="http://schemas.android.com/apk/res/android"
         xmlns:app="http://schemas.android.com/apk/res-auto"
+        xmlns:tools="http://schemas.android.com/tools"
+        tools:context=".view.activity.ContributorsActivity"
         >
 
     <data>
--- a/app/src/main/res/layout/activity_licenses.xml
+++ b/app/src/main/res/layout/activity_licenses.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <layout xmlns:android="http://schemas.android.com/apk/res/android"
         xmlns:app="http://schemas.android.com/apk/res-auto"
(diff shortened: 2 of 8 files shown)


konifar:
:eyes:

app/src/main/res/layout/activity_contributors.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <layout xmlns:android="http://schemas.android.com/apk/res/android"
         xmlns:app="http://schemas.android.com/apk/res-auto"
+        xmlns:tools="http://schemas.android.com/tools"
+        tools:context=".view.activity.ContributorsActivity"
```
konifar: I have a question 🙋 
I know this `tools:context` is set when I create Activity by AndroidStudio file template.
But to be honest, I don't know what is the benefit to specify `tools:context`.

ogapants: `tools:context` reflects the theme of Activity set in AndroidManifest on the Android Studio preview.
I thought that it was easier to understand the preview if it had this since this application also has a theme for each Activity.
As a result, the preview has not changed in the current themes, but it may be useful when the themes are changed in the future.
[The images of this site](http://tech.pjin.jp/blog/2013/06/26/android-%E3%80%8Cxmlnstools%E3%80%8D%E3%80%8Ctoolscontext%E3%80%8D%E3%81%A3%E3%81%A6%E3%81%AA%E3%82%93%E3%81%A0%E3%82%8D%E3%81%86%EF%BC%9F%E3%80%90android-tips%E3%80%91/) helps to understand.

(Also, I personally use it when jumping from the xml file to Activity)

If you feel unnecessary, I will just delete the mistake :D

konifar: Understood 📝 Thanks for explanation!
Actually, I always can not manage this `tools:context` 😢 So I wanted to know the benefit.

konifar:
LGTM! 💯 Thanks!

PR #324, opened 2017-02-12
Apply theme color for Dialog (Suggest)

ryugoo:
## Issue
- none

## Overview (Required)
- Apply theme color for Dialog
- Currently "AlertDialog" only used in "SettingsFragment"
  - This dialog radio button and cancel button are using the system default color (looks like a green)
- I suggest using a theme color

## Links
- https://material.io/guidelines/components/dialogs.html#dialogs-specs

> Dialog actions use system colors by default, but they should reflect your product's color palette. Use a contrasting color, such as the palette’s accent color, to distinguish dialog actions from dialog content.

Material design guidelines suggest using accent colors, but the accent color of the current theme is red. Since this looks like a "warning" color, I suggest a theme color.

If do not need the this PR, please close it!

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/res/values-v21/themes.xml
+++ b/app/src/main/res/values-v21/themes.xml
@@ -20,6 +20,8 @@
         <item name="android:statusBarColor">@android:color/transparent</item>
     </style>
 
-    <style name="DialogTheme" parent="Theme.AppCompat.Light.Dialog" />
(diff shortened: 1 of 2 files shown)


konifar:
:eyes:

konifar:
Thanks for good suggestion! I love it.
LGTM! 💯

PR #326, opened 2017-02-13
Remove unnecessary packagingOptions

shiraji:
## Issue
-

## Overview (Required)
~~There are 2 packagingOptions in `app/build.gradle`. Let's combine them.~~

~~BTW, do we really need `exclude 'LICENSE.txt'`???~~ 

Remove unnecessary `packagingOptions`

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -124,10 +124,6 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
 
-    packagingOptions {
-        exclude 'LICENSE.txt'
-    }
-
     jacoco {
         version "0.7.7.201606060606"
     }


konifar:
Thanks!

> BTW, do we really need exclude 'LICENSE.txt'

I can't remember why I added this 🤔  Maybe the build error occurred by some of library's LICENSE.txt.
If you can build without this `exclude` option, it's unnecessary.

shiraji:
> If you can build without this exclude option, it's unnecessary.

Let me check

shiraji:
Yea. In my machine, no error happens if I remove the setting.

Squashed the commits.

konifar:
Thanks! LGTM! 💯

PR #328, opened 2017-02-13
Register alarm when long clicking on session cell

gen0083:
## Issue
- Add my session from sessions page with long clicking session cell, alarm was not fired.

## Overview (Required)
-

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


konifar:
Good catch! ⚾️ LGTM! 👍

konifar:
Thanks for fix!

PR #334, opened 2017-02-14
Use Consumer as callback

tomoima525:
## Issue
- Use Consumer instead of Callback interface (to follow RxJava-ish style )

## Overview (Required)
- This PR is just an idea, I'll leave a decision to merge this PR up to @konifar 
- There are a Pros and Cons for using Consumer.

### Pros
- Reactive style, simple interface
- No `implements` required
- can use method reference (which is pretty neat in my opinion)

### Cons
- One consumer is responsible for one action ( interface style can handle as much as it want)
- Need to wrap with try-catch (which is a bit ugly)

## Links
NA

## Screenshot
NA

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/ContributorsFragment.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/ContributorsFragment.java
@@ -25,7 +25,7 @@ import io.github.droidkaigi.confsched2017.view.helper.IntentHelper;
 import io.github.droidkaigi.confsched2017.viewmodel.ContributorViewModel;
 import io.github.droidkaigi.confsched2017.viewmodel.ContributorsViewModel;
 
-public class ContributorsFragment extends BaseFragment implements ContributorsViewModel.Callback {
+public class ContributorsFragment extends BaseFragment {
 
     public static final String TAG = ContributorsFragment.class.getSimpleName();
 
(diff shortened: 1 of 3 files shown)


konifar:
Thanks for interesting idea 👍 
Currently, I prefer Callback style. But to be honest, I'm looking for the best way. One of my improvement idea is to inject Callback class to ViewModel.

Please let me consider till tonight (JST)! 📝

jmatsu:
How about using Lightweight-Stream-API's Consumer instead of Rx's Consumer?
It's not reactive x, but can remove one cons - `Need to wrap with try-catch (which is a bit ugly)`.

Anyway, I don't know what is the best way... and wanna see @konifar 's opinion. 😄

tomoima525:
Thank you for the comments! If you are good to go with this, I'll fix the build(looks like test is failing)

tomoima525:
> Lightweight-Stream-API's Consumer instead of Rx's Consumer

I never thought about it. Let me try that too.

konifar:
Sorry for late comment :bowing_man:
Finally, I want to adopt original Callback interface style.

To be honest, `Consumer` style you proposed is good as well :ok_woman:
I know the pros but I'm a bit negative for some cons.

---
1) Reactive style may not be friendly with all Androiders.
I think Reactive style is not familiar with all Androiders although `Consumer` is simple interface. I feel `.accept()` method is not Intuitive compared with `callback.onClickXX()`. I would like to make this project familiar to many Androiders if possible.

2) `Consumer#accept(T)` method has only one parameter.
If I want to pass some parameters, I have to use Pair or create a class.

3) One consumer is responsible for one action.
This is cons you mentioned. If ViewModel has 4 click event, it's necessary to add 4 `Consumer` class to ViewModel, right? I feel a bit complicated.

---
I think `Consumer` style is also good! This is almost a matter of taste 🍷

konifar:
@tomoima525 Sorry, can I close this? 🙇 I really appreciate your suggestion.

tomoima525:
@konifar Sorry for the late response. Ok, that's fine with me! Thanks for the comment 😉

PR #338, opened 2017-02-14
Stop restarting a process for service recreation

KeithYokoma:
- DebugOverlayService should live along with an app process and never be restarted again automatically after its death, causing unnecessary process restart 😭


konifar:
:eyes:

konifar:
LGTM 💯

PR #345, opened 2017-02-15
Add ChromeCustomTabs

shihochan:
## Issue
- none

## Overview (Required)
- I was going to add ChromeCustomTabs and fix to open web page.
- I'm super super sorry to do not use awesome [helper](https://github.com/DroidKaigi/conference-app-2017/blob/master/app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/IntentHelper.java) developed by our FRIEND. 😿  

## Links
- https://developer.chrome.com/multidevice/android/customtabs

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SponsorsFragment.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SponsorsFragment.java
@@ -1,9 +1,6 @@
 package io.github.droidkaigi.confsched2017.view.fragment;
 
-import com.annimon.stream.Optional;
-
 import android.content.Context;
-import android.content.Intent;
 import android.databinding.ObservableList;
 import android.graphics.Point;
 import android.os.Bundle;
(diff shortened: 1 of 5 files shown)


konifar:
:eyes:

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SponsorsFragment.java
```diff
@@ -81,7 +78,11 @@ private void initView() {
         binding.recyclerView.setLayoutManager(new LinearLayoutManager(getContext()));
     }
 
-    private static class SponsorAdapter
```
konifar: I guess it's not necessary to remove `static` 🤔

konifar:
Thanks! Commented!

> I'm super super sorry to do not use awesome helper developed by our FRIEND. 😿

Never mind.

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SponsorsFragment.java
shihochan: mmm... probably, I still need to study more.
static class can not access non-static method. It have to use non-static method `getActivity()` to open CustomTabs.
What is best practice to fix these cases?

konifar: Ah, sorry... I misunderstood 🙇 
Because this static class can call `getContext()`, but the method is defined in `ArrayRecyclerAdapter`.
As you said, it's impossible to call non-static method from static class.
Thanks for explanation

konifar:
LGTM! 💯 Thanks!

Issue #342, opened 2017-02-15
Never use new configuration to update the existing configuration


jmatsu:
## Overview (Required)
- Now, new instance of Configuration is used to update Locale. However, we just wanna update `only` Locale, so we should use the existing instance in this case.

## Links

- https://github.com/DroidKaigi/conference-app-2017/blob/995639ca51e8029bf0400b3b5d5540021f3d3fe0/app/src/main/java/io/github/droidkaigi/confsched2017/util/LocaleUtil.java#L39

jmatsu:
Thanks @kitwtnb !

PR #346, opened 2017-02-15
Fix configuration overwrite

kitwtnb:
## Issue
- #342 

## Overview (Required)
- Fixed it because configuration is overwrite.

## Links
- none


konifar:
LGTM! 💯 Thank you so much!

PR #347, opened 2017-02-15
Replace collect(Collectors.toList()) to Stream.toList().

operando:
## Issue
- none

## Overview (Required)
- Replace `collect(Collectors.toList())` to `Stream.toList()`.
- `Stream.toList()` which can be much faster and convenient than `collect(Collectors.toList())`!!
- This implementation does not call `collect(Collectors.toList())`, so it can be faster by reducing method calls.

## Links
- https://github.com/aNNiMON/Lightweight-Stream-API/releases/tag/v1.1.5
- http://static.javadoc.io/com.annimon/stream/1.1.5/com/annimon/stream/Stream.html#toList--


konifar:
:eyes:

konifar:
Cool! 🆒 I didn't know this! Thanks! 📝

konifar:
LGTM!!

PR #348, opened 2017-02-15
Implement a retry button for when contributors load failed

ogapants:
## Issue
none

## Overview (Required)
Loading will result in an error if you do not have a cache and can't connect to the Internet, but nothing changes on the screen.
In that case, I think that the application needs an approach such as showing a retry button.

I implemented it on ContributorsFragment for the time being, I think that the similar processing is necessary on other screens.
But since they are numerous, we will create an Issue if necessary 🙇 

## Links
none

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/res/layout/fragment_contributors.xml
+++ b/app/src/main/res/layout/fragment_contributors.xml
@@ -33,20 +33,22 @@
 
         </android.support.v4.widget.SwipeRefreshLayout>
 
-        <FrameLayout
-            android:layout_width="match_parent"
-            android:layout_height="match_parent"
-            android:background="@color/white"
+        <ProgressBar
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
             android:visibility="@{viewModel.loadingVisibility}"
-            tools:visibility="gone">
+            android:layout_gravity="center"
+            />
 
-            <ProgressBar
-                android:layout_width="wrap_content"
-                android:layout_height="wrap_content"
-                android:layout_gravity="center"
-                />
-
-        </FrameLayout>
+        <Button
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:onClick="@{viewModel::onClickRetry}"
+            android:text="@string/retry"
+            android:textColor="@color/white"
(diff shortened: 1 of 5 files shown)


konifar:
:eyes:

ogapants:
I thought now, should I show a Snackbar with retry action? 🤔

app/src/main/res/values/strings.xml
```diff
@@ -23,6 +23,7 @@
     <string name="map">Map</string>
     <string name="information">Information</string>
     <string name="settings">Settings</string>
+    <string name="retry">retry</string>
```
konifar: Can you change this to `Retry`?

app/src/main/res/layout/fragment_contributors.xml
```diff
@@ -33,20 +33,22 @@
...
 
-        </FrameLayout>
+        <Button
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:onClick="@{viewModel::onClickRetry}"
+            android:text="@string/retry"
+            android:textColor="@color/white"
+            android:visibility="@{viewModel.retryVisibility}"
```
konifar: Can you use `@style/Button`?

```java
<Button
    style="@style/Button"
    android:layout_gravity="center"
    android:onClick="@{viewModel::onClickRetry}"
    android:text="@string/retry"
    android:visibility="@{viewModel.retryVisibility}"
```

konifar:
> I thought now, should I show a Snackbar with retry action? 🤔

Hmm, I don't know which is better 🤔 But I prefer Snackbar.
Can you try? 🙇

ogapants:
Okay! I'll try it 💪

ogapants:
How about this? I feel that the code is simpler 🤔

konifar:
Looks good. I have one more order. Can you change the button text color? to white?

ogapants:
Yes, sir.
BTW, Should I move the error message to strings.xml?

konifar:
> Should I move the error message to strings.xml?

Yes, please!

ogapants:
I did it 😃

konifar:
LGTM! Thanks 👍

PR #351, opened 2017-02-16
Use injectable Navigator instead of callback of ViewModel

k-kagurazaka:
## Issue
None

## Overview (Required)
This is for suggestion, please close it if you are not a fun of this idea.

The goal of this PR is that remove all callbacks of `ViewModel` and uses injectable `Navigator`s instead.
Currently I completed following tasks:
- Rename `AppUtil` to `WebNavigator` and make it injectable.
- Use `WebNavigator` instead of callback.

I want to get reviews before working remaining tasks, so please don't merge this PR now.

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/BaseActivity.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/BaseActivity.java
@@ -25,6 +25,7 @@ import io.github.droidkaigi.confsched2017.MainApplication;
 import io.github.droidkaigi.confsched2017.R;
 import io.github.droidkaigi.confsched2017.di.ActivityComponent;
 import io.github.droidkaigi.confsched2017.di.ActivityModule;
+import io.github.droidkaigi.confsched2017.model.Session;
 import io.github.droidkaigi.confsched2017.view.helper.Navigator;
 import uk.co.chrisjenx.calligraphy.CalligraphyContextWrapper;
 
@@ -90,7 +91,32 @@ public abstract class BaseActivity extends AppCompatActivity implements Navigato
     }
 
     @Override
-    public void navigateToWebPage(@NonNull String url) {
+    public void navigateToSessionDetail(@NonNull Session session) {
+        raiseNavigationError();
+    }
+
+    @Override
+    public void navigateToFeedbackPage(@NonNull Session session) {
+        raiseNavigationError();
+    }
+
+    @Override
+    public void navigateToSponsorsPage() {
+        raiseNavigationError();
+    }
+
+    @Override
+    public void navigateToContributorsPage() {
+        raiseNavigationError();
+    }
+
+    @Override
+    public void navigateToLicensePage() {
+        raiseNavigationError();
+    }
+
+    @Override
+    public final void navigateToWebPage(@NonNull String url) {
         if (TextUtils.isEmpty(url) || !URLUtil.isNetworkUrl(url)) {
             return;
         }
@@ -101,6 +127,10 @@ public abstract class BaseActivity extends AppCompatActivity implements Navigato
                 .build();
 
         intent.launchUrl(this, Uri.parse(url));
+    }
+
+    private final void raiseNavigationError() {
+        throw new IllegalStateException("Couldn't navigate to the specified view.");
     }
 
     final void replaceFragment(@NonNull Fragment fragment, @IdRes @LayoutRes int layoutResId) {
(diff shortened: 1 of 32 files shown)


konifar:
Thanks! `WebNavigator` seems good.
I have a question! 🙋 
Do you have any idea other `Navigator` class for Activity/Dialog?

k-kagurazaka:
I have two plans:
1. Separate `Navigator` for each view, like `InformationViewNavigator`.

Pros.
- Developers cannot dispatch request that navigate to non-reachable view (e.g. dispatch navigateToSponsorsView on SessionsView).

Cons.
- We should create a lot of file (interfaces and Dagger module).

2. Create `ViewNavigator` that will be used for all transition.

Pros. and Cons. are reverse of plan 1.

Now I think plan 2 is better, what do you think about?

konifar:
@k-kagurazaka Thanks! I also think 2 is better since this app doesn't have so many views 😃

k-kagurazaka:
@konifar Thanks for your quick reply.
I will try to implement `ViewNavigator`.

k-kagurazaka:
I rewrite activity transition by injectable `Navigator`.
Currently some `Callback` interfaces exist yet, however we should apply MVVM pattern before removing these callbacks.
So I changed the focus of PR as changed title.

Would you like to review?

app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/BaseActivity.java
```diff
@@ -3,31 +3,40 @@
...
 import io.github.droidkaigi.confsched2017.MainApplication;
+import io.github.droidkaigi.confsched2017.R;
 import io.github.droidkaigi.confsched2017.di.ActivityComponent;
 import io.github.droidkaigi.confsched2017.di.ActivityModule;
+import io.github.droidkaigi.confsched2017.model.Session;
+import io.github.droidkaigi.confsched2017.view.helper.Navigator;
 import uk.co.chrisjenx.calligraphy.CalligraphyContextWrapper;
 
-public abstract class BaseActivity extends AppCompatActivity {
+public abstract class BaseActivity extends AppCompatActivity implements Navigator {
```
konifar: Hmm, I don't like `BaseActivity` implements `Navigator`. Because I want to keep `BaseActivity` small.
Please let me consider.

app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/BaseActivity.java
```diff
@@ -81,6 +90,49 @@ public boolean onKeyUp(int keyCode, KeyEvent event) {
...
+    @Override
+    public void navigateToContributorsPage() {
+        raiseNavigationError();
+    }
+
+    @Override
+    public void navigateToLicensePage() {
+        raiseNavigationError();
+    }
+
```
k-kagurazaka: All activity transition method throws an exception in `BaseActivity`.
So we can know the fact that we accidentally dispatch request which navigates to unreachable view by this exception.

app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/BaseActivity.java
k-kagurazaka: Of course we can define `NavigatorImpl` class that depends on `Activity`.
Would you like to fix?

konifar: Reasonable! Yes, It's better for me. Can you fix? 🙇

k-kagurazaka: If we extract `NavigatorImpl`, `Navigation` interface allows us to navigate all activities everywhere 😭

konifar: I know... `NavigatorImpl` way loses the pros of `BaseActivity implements Navigator` you mentioned.

> So we can know the fact that we accidentally dispatch request which navigates to unreachable view by this exception.

konifar: I don't like "God" class too :godmode: So I said "Please let me consider." 🙇

konifar: But currently, I think it's better to create `NavigatorImpl` than to implements `Navigator` in `BaseActivity` 💭

k-kagurazaka: OK.
I wait your decision 😇

k-kagurazaka: It looks like we've crisscrossed our comments.

I'm going to create non-final `Navigator` class that holding activity since interface is no longer needed.

k-kagurazaka:
I just extract `Navigator` from `BaseActivity` 😄

konifar:
LGTM! :cool: Thanks for great Refactoring! I learned a lot from this

PR #356, opened 2017-02-16
Avoid Jacoco warning

tnj:
Another quick fix: eliminate a build warning of Jacoco configuration.

## Issue
- None

## Overview (Required)
- I've just noticed that there's a build warning from Jacoco plugin.

```
It is no longer possible to set the Jacoco version in the jacoco {} block.
To update the version of Jacoco without updating the android plugin,
add a buildscript dependency on a newer version, for example: buildscript{    dependencies {
        classpath "org.jacoco:org.jacoco.core:0.7.5.201505241946"    }}
```

- I'm guessing @KeithYokoma should be aware of this
   - The warning message indicates that the current version specification of `0.7.7.201606060606` was not effective.
   - Though I'm not sure which version was used until now, this changeset might produce some differences in Jacoco result.

## Links
- https://docs.gradle.org/current/userguide/jacoco_plugin.html
   -  Looks like `version` property is no longer available


konifar:
:eyes:

konifar:
Thanks! Looks good 👍 
@KeithYokoma Can you review just in case 🙇

KeithYokoma:
Looks good!
[LGTM](https://lgtm.in/i/KLC2RkGbq)

[:+1:](https://lgtm.in/u/KLC2RkGbq)

konifar:
Amazing! 🐯

tnj:
Thanks! 👍

PR #363, opened 2017-02-18
Fix little grammar mistakes

hironytic:
## Issue
none

## Overview (Required)
- `are bind` -> `are bound`
- Change location of `or`

## Links
none

## Screenshot
none


konifar:
Good catch! LGTM
Thanks!

Issue #365, opened 2017-02-18
New object allocation in onDraw


eneim:
I wonder if [this line](https://github.com/DroidKaigi/conference-app-2017/blob/master/app/src/main/java/io/github/droidkaigi/confsched2017/view/customview/ParticlesAnimationView.java#L63) will allocate new Stream instances on every call. So it is something we don't want to do in onDraw. (This view is now used only in Plash screen so performance is not a big issue).

konifar:
Maybe there is no reason 💭

konifar:
@kitwtnb 
Thanks! #366

eneim:
@konifar Just made a comment on the PR :D.

PR #366, opened 2017-02-19
Using the 'for' statement instead of create new Stream instance.

kitwtnb:
## Issue
- https://github.com/DroidKaigi/conference-app-2017/issues/365

## Overview (Required)
- Create new Stream instance every time 'onDraw' is called.
- So using 'for' statement.

## Links
- none

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/customview/ParticlesAnimationView.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/customview/ParticlesAnimationView.java
@@ -60,8 +60,8 @@ public class ParticlesAnimationView extends View {
     @Override
     protected void onDraw(Canvas canvas) {
         super.onDraw(canvas);
-        Stream.of(particles).forEach(particle -> particle.draw(canvas, paint));
-        Stream.of(createLines(particles)).forEach(line -> line.draw(canvas, paint));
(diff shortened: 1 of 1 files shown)


konifar:
Thanks!! LGTM 💯

eneim:
@konifar As reference for [this blog post](https://publicobject.com/2016/12/19/seductive-code/), I suspect that this fix gives same result: a new instance of Iterator is **implicitly** allocated for each *iterator for loop* call. My suggestion is to use the for loop using index (which simply get the instance from the list reference, no new thing is allocated). Your code looks like 1990 java but ...

P/S: sorry for being a bit OCD >.<. But 60fps FTW :p.

konifar:
Sorry, Let me confirm your opinion 🤔 
You mean following code is better?

```java
for (int i = 0, size = particles.size(); i < size; i++) {
    particles.get(i).draw(canvas, paint);
}

List<Line> lines = createLines(particles);
for (int i = 0, size = lines.size(); i < size; i++) {
    lines.get(i).draw(canvas, paint);
}
```

eneim:
@konifar Yes. The point is to prevent object allocation in onDraw, and Iterator by ```for (Object item: items)``` will implicitly create one (this is compiler work, we can do nothing). 

The second part is to produce new lines for the particle lis, but I think is fine now, so just reduce the allocation as much as possible :D.

PR #367, opened 2017-02-19
Fix crash in splash when offline

ogapants:
## Issue
none

## Overview (Required)
Since the app has no error handling in `loadSessionsForCache()`, it crashes when fails to load sessions in splash. 💥 
I added only logs using Timber.
I think that we need to implement session reloading function in the future.

## Links
none

## Screenshot
none

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
@@ -1,11 +1,8 @@
 package io.github.droidkaigi.confsched2017.view.activity;
 
-import android.content.Context;
-import android.content.Intent;
 import android.databinding.DataBindingUtil;
 import android.os.Build;
 import android.os.Bundle;
-import android.os.Handler;
 import android.view.View;
 
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


konifar:
:one: :one:

app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
```diff
@@ -73,7 +71,8 @@ private void loadSessionsForCache() {
                     if (isFinishing()) return;
                     startActivity(MainActivity.createIntent(SplashActivity.this));
                 })
-                .subscribe();
+                .subscribe(objectObservable -> Timber.tag(TAG).d("Succeeded in loading sessions."),
```
konifar: Can you change valuable to `observable` from `objectObservable`?
In fact, lambda variable should be more short name like `o`, `t`.

konifar: `Succeeded to load sessions.` is better?

ogapants: I'll fix it soon! Thanks 🙇

ogapants: I think that "succeed to" means "引き継ぐ"... 🤔

konifar: Oh I didn't know that 🙇 Thanks

konifar:
LGTM! 💯

PR #368, opened 2017-02-19
Fix potential problems in ResourceResolver.loadJSONFromAsset()

sumio:
## Issue

`ResourceResolver.loadJSONFromAsset()` has the following code:

```java
            is = context.getAssets().open("json/" + jsonFileName);
            int size = is.available();
            byte[] buffer = new byte[size];
            is.read(buffer);
            json = new String(buffer, "UTF-8");
```

The above code may read only partial content of the file because `is.available()` may return less than the length of the file.

## Overview (Required)

- Use Apache Commons IO to avoid handling such a complicated stream behavior.

## Links

N/A

## Screenshot

N/A

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/ResourceResolver.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/ResourceResolver.java
@@ -1,7 +1,5 @@
 package io.github.droidkaigi.confsched2017.view.helper;
 
-import org.apache.commons.io.IOUtils;
-
 import android.content.Context;
 import android.support.annotation.StringRes;
 
@@ -11,6 +9,8 @@ import java.io.InputStream;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import okio.BufferedSource;
+import okio.Okio;
 import timber.log.Timber;
 
 @Singleton
@@ -35,14 +35,21 @@ public class ResourceResolver {
 
     public String loadJSONFromAsset(final String jsonFileName) {
         String json = null;
-        InputStream is = null;
+        BufferedSource bufferedSource = null;
         try {
-            is = context.getAssets().open("json/" + jsonFileName);
-            json = IOUtils.toString(is, "UTF-8");
+            InputStream is = context.getAssets().open("json/" + jsonFileName);
+            bufferedSource = Okio.buffer(Okio.source(is));
+            json = bufferedSource.readUtf8();
         } catch (IOException e) {
             Timber.tag(TAG).e(e, "assets/json/%s: read failed", jsonFileName);
         } finally {
-            IOUtils.closeQuietly(is);
+            try {
+                if (bufferedSource != null) {
+                    bufferedSource.close();
+                }
+            } catch (IOException e) {
+                // ignore
+            }
         }
         return json;
     }
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/ResourceResolver.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/ResourceResolver.java
@@ -1,5 +1,7 @@
 package io.github.droidkaigi.confsched2017.view.helper;
 
+import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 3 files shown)


eyedol:
@sumio good call on the use of `available()`. However, I think it's much better to use `Okio` to handle file reads as it's already a dependency. Something like the code below should work. I haven't read `Okio` source code yet but I believe it solves that issue. If not feel free to ignore this.

```java
String json = null;
BufferedSource source = null;
try {
    final InputStream is = context.getAssets().open("json/" + jsonFileName);
    source = Okio.buffer(Okio.source(is));
    json = source.readUtf8();
} catch (IOException e) {
    Timber.tag(TAG).e(e, "assets/json/%s: read failed", jsonFileName);
} finally {
    try {
        if (source != null) {
            source.close();
        }
    } catch (IOException e) {
        Timber.tag(TAG).e(e, "assets/json/%s: read failed", jsonFileName);
    }
}
return json;       
```

konifar:
:eyes:

sumio:
@eyedol Thanks for your nice suggestion!
I'll rewrite it by using Okio in this night.

sumio:
Rewriting is done!
commons-io dependecy is removed.

sumio:
It seems that circleCI is failed in `ContributorsLocalDataSourceTest > updateAllAsyncAsInsert` which is unrelated in this PR...
(and all tests are passed in my local environment)

konifar:
@sumio Sorry, CircleCI is not stable now 🙇

sumio:
@konifar Thanks for your reply.
OK. I won't mind the result.

app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/ResourceResolver.java
```diff
@@ -33,22 +35,20 @@ public String getString(@StringRes int resId, Object... formatArgs) {
...
-        InputStream is = null;
+        BufferedSource bufferedSource = null;
         try {
-            is = context.getAssets().open("json/" + jsonFileName);
-            int size = is.available();
-            byte[] buffer = new byte[size];
-            is.read(buffer);
-            json = new String(buffer, "UTF-8");
+            InputStream is = context.getAssets().open("json/" + jsonFileName);
+            bufferedSource = Okio.buffer(Okio.source(is));
```
konifar: 🆒

app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/ResourceResolver.java
```diff
@@ -33,22 +35,20 @@ public String getString(@StringRes int resId, Object... formatArgs) {
...
         } finally {
             try {
-                if (is != null) {
-                    is.close();
+                if (bufferedSource != null) {
+                    bufferedSource.close();
                 }
             } catch (IOException e) {
-                Timber.e(TAG, e.getMessage(), e);
+                // ignore
```
konifar: 🙋 Why did you ignore here?

konifar:
@sumio Sorry for late review  🙇 LGTM! Thanks for contribution! 💯 
@eyedol Thanks for great suggestion! I appreciate you!

konifar:
Hope CI success 🙏

konifar:
👏

app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/ResourceResolver.java
sumio: Personally I think, it is commonly used that failure in finally block is ignored quietly.
(cf. [IOUtils.closeQuietly](https://commons.apache.org/proper/commons-io/javadocs/api-release/org/apache/commons/io/IOUtils.html#closeQuietly%28java.io.Closeable%29))

That said, it should be respected that project owners' opinion. 
Shall I put the log?

konifar: Understood! 📝 I follow your opinion. Thanks for IOUtils example!

PR #375, opened 2017-02-21
Feature/particles view update

eneim:
## IMPORTANT: 
- ~~I commented out some lines in SplashActivity to make the sample, please read it and un-comment for normal behaviour.~~
- I'm using AS RC-1 so please manually change gradle plugin back to 2.2.3, or I will do it after the review.

## Issue
- Current ```ParticlesAnimationView``` is beautifully implemented. But it comes with some trade off as follow:
  - Redundant Object allocation inside onDraw method: ```createLines(particles)``` uses Stream which will allocate new Iterators each calls. Furthermore it unnecessarily generates new Lines.
  - "Random" should be used once only to ensure the randomness of number. See its doc for more detail.
  - Current implementation of ```createLines(particles)``` will draw each line twice, which is not so good.
  - Calling ```drawLines``` to draw a single line is not so good too.
  - Private classes can be Private static class, to prevent leak (not now, just in case).

## Overview: what I did
- Make private class (Particle, Line) static.
- Re-implement Line class, fix some step in Particle class that would allocate new objects in onDraw().
- Cache necessary objects as ParticlesAnimationView's member, and just update them if need (see Random instance, List<Line> instance for example).
  - This fix reduce the number of Line instance from few thousands (数万 lines in some cases) to 780 (= C(40, 2)) (see my comment).
- Improve onDraw and so on. Stream is no longer used here ...
- ```isShouldLinked``` is actually ```shouldBeLinked``` I think :D.

## Discussion: so it worth the change?
- Current View is shown in few seconds, so there may be no need to do redundant work.
- But ... some may want to use it as Cover background, or Title view etc, so improving it here can help.
- See screenshot for more information (see number of GC calls, magenta (or blue?) in GPU log ...) (sorry don't know how to align 2 screenshot horizontally, please help).
## Screenshot
| Before | After |
| :--: | :--: |
| screen shot 2017-02-21 at 23 55 34 | 
screen shot 2017-02-21 at 23 59 15 |

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -226,7 +226,7 @@ dependencies {
     debugCompile 'com.tomoima.debot:debot:2.0.2'
     releaseCompile 'com.tomoima.debot:debot-no-op:2.0.2'
 
-    // fps
+    // fps, track the frame rate of animation view.
     compile 'jp.wasabeef:takt:1.0.3'
 
     // Test
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -226,6 +226,9 @@ dependencies {
     debugCompile 'com.tomoima.debot:debot:2.0.2'
     releaseCompile 'com.tomoima.debot:debot-no-op:2.0.2'
 
+    // fps
+    compile 'jp.wasabeef:takt:1.0.3'
+
     // Test
     testCompile 'junit:junit:4.12'
     testCompile "org.robolectric:robolectric:3.2.2"
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
@@ -54,7 +54,10 @@ public class SplashActivity extends BaseActivity {
     @Override
     protected void onStart() {
         super.onStart();
+        // TODO !IMPORTANT Uncomment this line or you will never see the schedule ...
         // loadSessionsForCache();
+
+        // Starting new Activity normally will not destroy this Activity, so set this up in start/stop cycle
         Takt.stock(getApplication())
                 .seat(Seat.BOTTOM_RIGHT)
                 .interval(250)
@@ -65,7 +68,10 @@ public class SplashActivity extends BaseActivity {
     @Override
     protected void onStop() {
         super.onStop();
+        // TODO !IMPORTANT If you uncommented that line, also uncomment this line or else LeakCanary will spam you.
         // compositeDisposable.dispose();
+
+        // Stop tracking the frame rate.
         Takt.finish();
     }
 
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
@@ -56,7 +56,6 @@ public class SplashActivity extends BaseActivity {
         loadSessionsForCache();
 
         // Starting new Activity normally will not destroy this Activity, so set this up in start/stop cycle
-        // TODO remove it if you hate it ...
         Takt.stock(getApplication())
                 .seat(Seat.BOTTOM_RIGHT)
                 .interval(250)
@@ -70,7 +69,6 @@ public class SplashActivity extends BaseActivity {
         compositeDisposable.dispose();
 
         // Stop tracking the frame rate.
-        // TODO remove it if you hate it ...
         Takt.finish();
     }
 
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
@@ -3,6 +3,7 @@ package io.github.droidkaigi.confsched2017.view.activity;
 import android.databinding.DataBindingUtil;
 import android.os.Build;
 import android.os.Bundle;
+import android.util.Log;
 import android.view.View;
 
 import java.util.Locale;
@@ -19,6 +20,8 @@ import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.disposables.CompositeDisposable;
 import io.reactivex.disposables.Disposable;
 import io.reactivex.schedulers.Schedulers;
+import jp.wasabeef.takt.Seat;
+import jp.wasabeef.takt.Takt;
 import timber.log.Timber;
 
 public class SplashActivity extends BaseActivity {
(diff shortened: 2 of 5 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
```diff
@@ -51,13 +54,25 @@ protected void onCreate(Bundle savedInstanceState) {
...
         super.onStart();
-        loadSessionsForCache();
+        // TODO !IMPORTANT Uncomment this line or you will never see the schedule ...
+        // loadSessionsForCache();
+
+        // Starting new Activity normally will not destroy this Activity, so set this up in start/stop cycle
+        Takt.stock(getApplication())
+                .seat(Seat.BOTTOM_RIGHT)
+                .interval(250)
+                .listener(fps -> Log.w("DroidKaigi", "heartbeat() called with: fps = [" + fps + "]"))
```
konifar: **[Android Lint]** <sub>reported by [reviewdog](https://github.com/haya14busa/reviewdog) :dog:</sub>
Error: Using 'Log' instead of 'Timber' [LogNotTimber]

konifar:
Thank you for contribution! Interesting!

eneim:
Thanks. Let me fix the license and conflict in few minutes :D

eneim:
FYI @konifar I would like to ask @roana0229 (author of this View) to review this if possible.

app/build.gradle
```diff
@@ -226,6 +226,9 @@ dependencies {
     debugCompile 'com.tomoima.debot:debot:2.0.2'
     releaseCompile 'com.tomoima.debot:debot-no-op:2.0.2'
 
+    // fps, track the frame rate of animation view.
+    compile 'jp.wasabeef:takt:1.0.3'
```
konifar: 🆒

app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
```diff
@@ -52,12 +54,24 @@ protected void onCreate(Bundle savedInstanceState) {
     protected void onStart() {
         super.onStart();
         loadSessionsForCache();
+
+        // Starting new Activity normally will not destroy this Activity, so set this up in start/stop cycle
+        // TODO remove it if you hate it ...
```
konifar: I like it. So let's remove this `TODO` comment.

app/src/main/java/io/github/droidkaigi/confsched2017/view/customview/ParticlesAnimationView.java
```diff
@@ -100,68 +122,43 @@ public Shader resize(int width, int height) {
...
-                        .forEach(line -> {
-                            if (!lines.contains(line)) lines.add(line);
-                        })
-                );
-        return lines;
-    }
-
-    private class Line {
+    /**
+     * A Pair of 'Particles' whose centers can be 'linked to each other ... Called 'Line'.
```
konifar: 👍

konifar:
Looks good. Thanks 😃 
@roana0229 Can you review this?

app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
eneim: Just push a commit to remove the TODO :D.

eneim:
@konifar my latest commit cause a failed build on circleci, checking it, turns out ```:app:testDevelopDebugUnitTest``` failed with one case. But running it on my machine passes all tests. Can you manually trigger a rebuild? If it keeps failing again I will try to figure out what happened.

konifar:
Sorry, currently CI is not stable. I'll work on it tonight.

eneim:
Thanks, just take your time :D

app/src/main/java/io/github/droidkaigi/confsched2017/view/customview/ParticlesAnimationView.java
```diff
@@ -100,68 +122,43 @@ public Shader resize(int width, int height) {
...
+            super(first, second);
         }
 
-        public void draw(Canvas canvas, Paint paint) {
-            final double distance = Math.floor(Math.sqrt(
-                    (point[2] - point[0]) * (point[2] - point[0])
-                            + (point[3] - point[1]) * (point[3] - point[1])
-            ));
+        void draw(Canvas canvas, Paint paint) {
+            if (!first.shouldBeLinked(LINK_HEXAGON_DISTANCE, second)) {
```
roana0229: In my opinion, `shouldBeLinked` is not necessary if first particle compare second particle in `draw`.

eneim: @roana0229 not really get your full point, but do you mean to calling ```shouldBeLinked``` from ```ParticlesAnimationView#onDraw``` instead of here? It will end up removing the need of using "Line" object (a Line is actually just a bridge between 2 particles :D).

Yesterday I thought about something like:

```java
    @Override
    protected void onDraw(Canvas canvas) {
        super.onDraw(canvas);
        for (int i = 0, size = particles.size(); i < size; i++) {
            particles.get(i).draw(canvas, paint);
        }

        // for (int i = 0, size = lines.size(); i < size; i++) {
        //    lines.get(i).draw(canvas, paint);
        //}

        for (int i = 0; i < particles.size(); i++) {
            for (int j = 0; j < particles.size(); j++) {
                if (particles.get(i).shouldBeLinked(LINK_HEXAGON_DISTANCE, particles.get(j))) {
                    // TODO draw, use the code from Line
                }
            }
        }
    }
```

I think what you mean is the following:

```java
    // ParticlesAnimationView
    @Override
    protected void onDraw(Canvas canvas) {
        super.onDraw(canvas);
        for (int i = 0, size = particles.size(); i < size; i++) {
            particles.get(i).draw(canvas, paint);
        }

         for (int i = 0, size = lines.size(); i < size; i++) {
             if (lines.get(i).shouldDraw()) {
                 lines.get(i).draw(canvas, paint);
             }
        }
    }

    // ParticlesAnimationView$Line
    ///// shouldDraw simply returns the comparison
    boolean shouldDraw() {
        return first.shouldBeLinked(LINK_HEXAGON_DISTANCE, second);
    }
```

Either of above gain any huge impact (but the later looks clean and better, in fact :D). For now I would like to keep current version (this PR is pretty huge now) and maybe another PR from you if you really want to :p. So it is up to you ^^.

roana0229: I thought that `the following code` was good.

> For now I would like to keep current version (this PR is pretty huge now) and maybe another PR from you

OK! 😄

roana0229:
LGTM 👍

konifar:
Thanks for improvement!

PR #385, opened 2017-02-24
Fixed #383

henteko:
## Issue
- #383 

## Overview (Required)
- Show icon of speakers 🎉

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

PR #386, opened 2017-02-24
Refactored speaker image url binding

konifar:
## Issue
https://github.com/DroidKaigi/conference-app-2017/pull/385

## Overview (Required)
- Refactored speaker image url binding.
- Originally the image url is converted in `DataBindingHelper` class.
- But I think this should be converted in `Speaker` class.

## Screenshot
I'll upload later.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -45,6 +45,8 @@ android {
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
 
         buildConfigField "String", "CONFERENCE_TIMEZONE", "\"Asia/Tokyo\""
+        buildConfigField "String", "API_ROOT", "\"https://droidkaigi.github.io/2017/\""
+        buildConfigField "String", "STATIC_ROOT", "\"https://droidkaigi.github.io/2017\""
         buildConfigField "String", "GIT_SHA", "\"${gitSha}\""
         buildConfigField "String", "GIT_BRANCH", "\"${gitBranch}\""
         buildConfigField "String", "BUILD_TIME", "\"${buildTime}\""
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/model/Speaker.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/model/Speaker.java
@@ -51,7 +51,7 @@ public class Speaker {
         } else if (imageUrl.startsWith("/")) {
             return BuildConfig.STATIC_ROOT + imageUrl;
         } else {
-            Timber.tag(TAG).e("Invalid image url: ", imageUrl);
+            Timber.tag(TAG).e("Invalid image url: %s", imageUrl);
             return null;
         }
     }
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/model/Speaker.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/model/Speaker.java
@@ -8,8 +8,13 @@ import com.github.gfx.android.orma.annotation.Table;
 
 import android.support.annotation.Nullable;
 
+import io.github.droidkaigi.confsched2017.BuildConfig;
+import timber.log.Timber;
+
 @Table
 public class Speaker {
+
+    private static final String TAG = Speaker.class.getSimpleName();
 
     @PrimaryKey(auto = false)
     @Column(indexed = true)
@@ -34,4 +39,20 @@ public class Speaker {
     @Nullable
     @SerializedName("github_name")
     public String githubName;
+
+    @Nullable
+    public String getAdjustedImageUrl() {
+        if (imageUrl == null) {
+            return null;
+        }
+
+        if (imageUrl.startsWith("http://") || imageUrl.startsWith("https://")) {
+            return imageUrl;
+        } else if (imageUrl.startsWith("/")) {
+            return BuildConfig.STATIC_ROOT + imageUrl;
+        } else {
+            Timber.tag(TAG).e("Invalid image url: ", imageUrl);
+            return null;
+        }
+    }
 }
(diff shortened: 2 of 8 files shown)


app/src/main/java/io/github/droidkaigi/confsched2017/view/helper/DataBindingHelper.java
```diff
@@ -68,7 +68,7 @@ private static void setImageUrlWithSize(ImageView imageView, @Nullable String im
             final int size = Math.round(sizeInDimen);
             imageView.setBackground(ContextCompat.getDrawable(imageView.getContext(), R.drawable.circle_border_grey200));
             Picasso.with(imageView.getContext())
-                    .load(convertToFullImageUrl(imageUrl))
```
konifar: This image url should be converted in ViewModel class.
Plus this method is called from `setPhotoImageUrl()`, so it's not good to convert here.

app/src/main/java/io/github/droidkaigi/confsched2017/viewmodel/SessionDetailViewModel.java
```diff
@@ -114,6 +121,21 @@ public void destroy() {
...
 
+    @Nullable
+    private String createSpeakerImageUrl(@Nullable Speaker speaker) {
+        if (speaker == null) {
+            return null;
+        }
+
+        String imageUrl = speaker.imageUrl;
+
+        if (!TextUtils.isEmpty(imageUrl) && imageUrl.startsWith("/")) {
```
konifar: `imageUrl.startsWith("/")` is not necessary? 🤔

k-kagurazaka: Some speakers have image_url that isn't start with / (like https://...) , so we shoud handle url type or fix sessions.json.

konifar: Oh, thanks!
I will fix the JSON.

app/src/main/java/io/github/droidkaigi/confsched2017/model/Speaker.java
```diff
@@ -34,4 +39,20 @@
...
     @SerializedName("github_name")
     public String githubName;
+
+    @Nullable
+    public String getAdjustedImageUrl() {
+        if (imageUrl == null) {
+            return null;
+        }
+
+        if (imageUrl.startsWith("http://") || imageUrl.startsWith("https://")) {
```
konifar: To be honest, I want to use `URLUtil.isNetworkUrl(imageUrl)`, but it depends on Android code.

app/src/main/java/io/github/droidkaigi/confsched2017/model/Speaker.java
```diff
@@ -34,4 +39,20 @@
...
+            return imageUrl;
+        } else if (imageUrl.startsWith("/")) {
+            return BuildConfig.STATIC_ROOT + imageUrl;
+        } else {
+            Timber.tag(TAG).e("Invalid image url: ", imageUrl);
```
konifar: **[Android Lint]** <sub>reported by [reviewdog](https://github.com/haya14busa/reviewdog) :dog:</sub>
Error: Wrong argument count, format string Invalid image url:  requires 0 but format call supplies 1 [TimberArgCount]

konifar: Thanks dog

konifar:
Can anyone review this? 🙇

app/build.gradle
```diff
@@ -45,6 +45,8 @@ android {
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
 
         buildConfigField "String", "CONFERENCE_TIMEZONE", "\"Asia/Tokyo\""
+        buildConfigField "String", "API_ROOT", "\"https://droidkaigi.github.io/2017/\""
```
konifar: Why did I divide `API_ROOT` and `STATIC_ROOT` though both are similar?
Because retrofit base url have to contain `/` on the end, and the static root url may be changed in the future.

app/src/test/java/io/github/droidkaigi/confsched2017/model/SpeakerTest.kt
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Test
+    @Throws(Exception::class)
+    fun getAdjustedImageUrlWhenInvalidPath() {
+        val url = "images/1.jpg"
+        val speaker = Speaker().apply { imageUrl = url }
+        speaker.adjustedImageUrl.should be null
+    }
+}
```
k-kagurazaka: Great! 👍

app/src/main/java/io/github/droidkaigi/confsched2017/model/Speaker.java
k-kagurazaka: In my opinion, I want to keep models independent as far as possible.
However in this case, we can use `URLUtil` since it's a simple utility class.
It's a matter of taste.

konifar: @k-kagurazaka Thanks! I know what you mean. But if I use `URLUtil` class, I also use Robolectric in the test code. As you said, it's not big problem, but I want to avoid it.

k-kagurazaka:
LGTM.
I'm a fun of your idea that moves image url logic to model (non-anemic) 💕

konifar:
Thanks for quick review! Appreciate you

app/src/main/java/io/github/droidkaigi/confsched2017/model/Speaker.java
k-kagurazaka: I haven't thought through that far.
I agree with your choice.

konifar:
Let me merge

Issue #377, opened 2017-02-22
User is able to return to Splash screen when tapping on back button immediately after Sessions screen is launched


miga9:
## Overview (Required)
- User is able to return to Splash screen when tapping on back button immediately after Sessions screen is launched. 

It does not happen after a few seconds are elapsed after Sessions screen is launched. App closes fine after tapping on back button again.

## Reproduction steps
- Launch the app.
- Splash screen loads.
- Sessions screen is opened. Tap on back button within a couple seconds.
- Observe that user is able to return to Splash screen.

## Links
- N/A

## Screenshot
<img src="" width="300" />

konifar:
Thanks shiraji!

PR #387, opened 2017-02-24
Finish splash activity manually

shiraji:
## Issue
#377

## Overview (Required)
It seems `android:noHistory="true"` doesn't finish the activity if user tap back button right after launching new activity.

## Links
I couldn't find this behavior's documentation. 

## Question

Could anyone point me if there is other concern?

## What I did

I tried the following code 

```java
    private void loadSessionsForCache() {
//        Disposable disposable = Single.zip(sessionsRepository.findAll(Locale.getDefault()),
//                mySessionsRepository.findAll(),
//                Single.timer(MINIMUM_LOADING_TIME, TimeUnit.MILLISECONDS),
//                (sessions, mySessions, obj) -> Observable.empty())
//                .subscribeOn(Schedulers.io())
//                .observeOn(AndroidSchedulers.mainThread())
//                .doFinally(() -> {
//                    if (isFinishing()) return;
//                    startActivity(MainActivity.createIntent(SplashActivity.this));
//                    SplashActivity.this.finish();
//                })
//                .subscribe(observable -> Timber.tag(TAG).d("Succeeded in loading sessions."),
//                        throwable -> Timber.tag(TAG).e(throwable, "Failed to load sessions."));
//        compositeDisposable.add(disposable);

        startActivity(MainActivity.createIntent(SplashActivity.this));
    }
```

This become no Splash screen but it still show me Splash screen if I tap back button just after launching session screen. This means Rx has nothing wrong with this behavior.

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
+++ b/app/src/main/java/io/github/droidkaigi/confsched2017/view/activity/SplashActivity.java
@@ -70,6 +70,7 @@ public class SplashActivity extends BaseActivity {
                 .doFinally(() -> {
                     if (isFinishing()) return;
                     startActivity(MainActivity.createIntent(SplashActivity.this));
+                    SplashActivity.this.finish();
                 })
                 .subscribe(observable -> Timber.tag(TAG).d("Succeeded in loading sessions."),
                         throwable -> Timber.tag(TAG).e(throwable, "Failed to load sessions."));


konifar:
:eyes:

eneim:
@shiraji About your question, it simply immediately starts new Activity, so you don't see the Splash screen. In fact it has the same behaviour. You can dump the back stack to see which activities are staying around :D

shiraji:
Sorry for confusing you.  I couldn't find `android:noHistory="true"` behavior's documentation.

konifar:
Hmm, it's weird...
I tried to add `android:launchMode="singleTask"` to MainActivity, but same problem occurred.
`android:noHistory="true"` seems not to work well.

But anyway your workaround works. Thanks

konifar:
LGTM!

PR #362, opened 2017-02-18
Add tests for ContributorsRepository and data sources

k-kagurazaka:
## Issue
None

## Overview (Required)
- Add more tests to `ContributorsRepositoryTest`
- Add `ContributorsLocalDataSourceTest` and `ContributorsRemoteDataSourceTest`

PR #389, opened 2017-02-25
Ignore unstable test

k-kagurazaka:
## Issue
None

## Overview (Required)
- Ignore the unstable test `ContributorsLocalDataSourceTest#updateAllAsyncAsInsert` which is added in #362 by me 🙇 🙇 

I think this problem is caused by scheduler operation of Rx, but I cannot solve yet since this test always passes in my local env.
Of course the best scenario is that makes this test stable, but now I just added `@Ignore` annotation 😭


konifar:
Sorry... I should have done this 🙇

konifar:
LGTM!

PR #406, opened 2017-03-07
Hide developer menu on production build

k-kagurazaka:
## Issue
- None

## Overview (Required)
- Hide developer menu on production build.

## Screenshot
Before | After
:--: | :--:
 |


konifar:
Thanks! But actually, this developer option will be talked in some sessions at DroidKaigi.
That's why I didn't hide this menu.
So please let me close this 🙇

k-kagurazaka:
Alright, I got it!

Issue #425, opened 2017-03-10
Crush on SessionDetailFragment


kwmt:
## Overview (Required)
- When language setting is change on SessionDetailFragment (activity is recreated), NPE occurs.

```
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.app.Activity.setTheme(int)' on a null object reference
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at io.github.droidkaigi.confsched2017.view.fragment.SessionDetailFragment.initTheme(SourceFile:72)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at io.github.droidkaigi.confsched2017.view.fragment.SessionDetailFragment.lambda$onCreateView$0(SourceFile:101)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at io.github.droidkaigi.confsched2017.view.fragment.SessionDetailFragment$$Lambda$1.run(Unknown Source)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at io.reactivex.internal.observers.CallbackCompletableObserver.onComplete(SourceFile:52)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at io.reactivex.internal.operators.completable.CompletableObserveOn$ObserveOnCompletableObserver.run(SourceFile:90)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at io.reactivex.android.schedulers.HandlerScheduler$ScheduledRunnable.run(SourceFile:109)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at android.os.Handler.handleCallback(Handler.java:739)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at android.os.Handler.dispatchMessage(Handler.java:95)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at android.os.Looper.loop(Looper.java:148)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at android.app.ActivityThread.main(ActivityThread.java:5417)
03-10 09:55:28.344 27594 27594 E SessionDetailFragment: 	at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #427, opened 2017-03-10
Fixed because NPE occurred, when activity is recreated.

kwmt:
## Issue
- #425 

## Overview (Required)
- NPE occurred, when activity is recreated.

## Links
- http://ja.stackoverflow.com/a/6196

## Screenshot
Before | After
:--: | :--:
 |


app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionDetailFragment.java
```diff
@@ -64,8 +64,8 @@ public void onStop() {
         compositeDisposable.dispose();
     }
 
-    private void initTheme() {
-        Activity activity = getActivity();
+    private void initTheme(Activity activity) {
+        if (activity == null) { return; }
```
konifar: Actually, this is not aligned our format.
But it's not necessary to fix! 👍

konifar:
LGTM! 💯

PR #430, opened 2017-03-10
Scroll to the day of the current day.

sys1yagi:
## Issue

None

## Overview (Required)

If today matches the opening date, scroll to that day.

## Links
-

## Screenshot
Before | After
:--: | :--:
 | a


app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
```diff
@@ -217,6 +222,26 @@ private void renderSessions(List<SessionViewModel> adjustedSessionViewModels) {
...
+                .filter(pair -> todayString.equals(pair.getSecond()))
+                .findFirst().orElse(new IntPair<>(0, "")).getFirst();
+    }
+
+    private void scrollTo(int position) {
+        binding.recyclerView.getViewTreeObserver().addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
+            @Override
+            public void onGlobalLayout() {
+                binding.recyclerView.getLayoutManager().scrollToPosition(position);
+                binding.recyclerView.getViewTreeObserver().removeOnGlobalLayoutListener(this);
```
sys1yagi: Is there a better way? 🤔

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
```diff
@@ -217,6 +222,28 @@ private void renderSessions(List<SessionViewModel> adjustedSessionViewModels) {
...
+    }
+
+    private int extractTodayPos(Context context, Date today, List<SessionViewModel> adjustedSessionViewModels) {
+        String todayString = DateUtil.getMonthDate(today, context);
+        return Stream.of(adjustedSessionViewModels)
+                .map(SessionViewModel::getFormattedDate)
+                .indexed()
+                .filter(pair -> todayString.equals(pair.getSecond()))
+                .findFirst()
+                .orElse(new IntPair<>(0, ""))
```
sys1yagi: If there is no match, position 0 is returned.

konifar:
:eyes:

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
konifar: Is it better to use `Stream`? In this case, just for loop seems easy. What do you think?

```java
for (int i = 0; i<= adjustedSessionViewModels.size(); i++) {
  SessionViewModel viewModel = adjustedSessionViewModels.get(i);
  if (todayString.equals(viewModel.getFormattedDate()) {
    return i;
  }
  return 0;
}
```

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
konifar: Sorry, I don't know why `OnGlobalLayoutListener` is needed 🙇 
Is it enough just to call `binding.recyclerView.getLayoutManager().scrollToPosition(position);` ?

app/src/main/java/io/github/droidkaigi/confsched2017/view/fragment/SessionsFragment.java
sys1yagi: agreeee~

